PR #3, opened 2022-01-31
Implement Plurimath Gem that bridges the latexmath and asciimath gems

### LINKED ISSUE #2: Implement Plurimath gem that bridges the latexmath and asciimath gems
The Plurimath gem is supposed to provide a central interface for conversion of multiple math representation languages, including:

* MathML presentation
* MathML semantical
* Microsoft Office Math Markup Language ([omml2mathml](https://github.com/plurimath/omml2mathml))
* AsciiMath ([mathml2asciimath](https://github.com/plurimath/mathml2asciimath))
* LaTeX math ([latexmath](https://github.com/plurimath/latexmath))
* HTML math ([html2asciimath](https://github.com/plurimath/html2asciimath))

Right now we have individual gems that handle 1-to-1 conversions, but we want to combine it all using a single math representation model.

That model may likely be MathML 3.0.

It should also handle UnitsML syntax.

cc: @hassanakbar

### LINKED ISSUE #6: Implement base structure of Plurimath gem
Create classes in `Plurimath::Math` and `Plurimath::Math::Formula` namespaces for conversions

suleman-uzair:
Converts from
1. Asciimath -> Unitsml
2. Html -> Asciimath
3. Latex -> Mathml
4. Mathml -> Asciimath
5. Omml -> Mathml
6. Unicode -> Latex

Closes #6


Gemfile
```diff
@@ -5,3 +5,4 @@ gemspec
 
 gem "rake", "~> 12.0"
 gem "rspec", "~> 3.0"
+gem 'html2asciimath', git: 'git@github.com:plurimath/html2asciimath.git'
```
HassanAkbar, line 8: @ronaldtse We need to publish this gem.
Let me know if I need to get it done?

ronaldtse: This is now published. Thanks!

ronaldtse: This has been just published now.

lib/plurimath/converters/omml.rb
```diff
@@ -0,0 +1,31 @@
...
+# </html>
+class Omml
+  attr_accessor :text
+
+  def initialize(str)
+    @text = str
+  end
+
+  def to_mathml
+    response = Omml2Mathml.convert(text)
```
HassanAkbar, line 28: @ronaldtse `Omml2Mathml` currently receives a filename as an input while every other gem takes a string.

To make things consistent, can we modify its input to a string, put it in a temp file convert it, and return the string?

ronaldtse:
The key point here is to create classes for every math data element used in a MathML formula, such as `mo`, `mfenced`, etc. You can find the full list here:
https://developer.mozilla.org/en-US/docs/Web/MathML/Element

suleman-uzair:
> The key point here is to create classes for every math data element used in a MathML formula, such as `mo`, `mfenced`, etc. You can find the full list here: https://developer.mozilla.org/en-US/docs/Web/MathML/Element

@ronaldtse We need to transform the text/strings into Ruby objects of Mathml (such as `mo` `mfenced` and all other elements in this [link](https://developer.mozilla.org/en-US/docs/Web/MathML/Element)) ?
As part of this issue, we want to create classes for each Mathml element and use them ?
Let me know if I am missing something here.

suleman-uzair:
@ronaldtse Should we add support of Ruby 2.4 and above in [html2asciimath](https://github.com/plurimath/html2asciimath/pull/4) gem OR remove support of Ruby version 2.4/2.5 or less from [plurimath](https://github.com/plurimath/plurimath/)?

ronaldtse:
@suleman-uzair let's settle on Ruby 2.6 being the last supported version since 2.5 is already EOLed. Thanks.

suleman-uzair:
@ronaldtse how do we intend to use these classes in this gem?

ronaldtse:
We want something like this to work:

```ruby
formula = Plurimath.parse('sum_(i=1)^n (i + 100)', :asciimath)
formula.class
#=> Plurimath::Formula <...>
formula.to_mathml.to_s
#=> '<math ns= ...> ... </math>`
formula.root
#=> Plurimath::Math::Summation(:lower => Plurimath::Math::Number<1>, :upper => Plurimath::Math::Variable<n>, :content => Plurimath::Math::) ...

def print_children(node)
  node.children.each do |child|
    puts "LaTeX representation: #{child.to_latex_math.to_s}"
    puts "AsciiMath representation: #{child.to_asciimath.to_s}"
    puts "MathML representation: #{child.to_mathml.to_s}"
    print_children(node) unless node.children.size == 0
  end
end
```

HassanAkbar:
@ronaldtse From what I understand we need to add classes not only for `MathML` but for every representation that we are handling so, we can store it in the parsed state.

Is this correct or am I missing something here?

ronaldtse:
@HassanAkbar your understanding is correct indeed 👍  The conversion would basically be a parse-tree to parse-tree conversion.

HassanAkbar:
@ronaldtse We should first create all the parsers to parse the input into a `Plurimath::Formula` and then in the next step we should convert the `Plurimath::Formula` into different representations.

We can use gems if exists, like [asciimath](https://github.com/asciidoctor/asciimath) for the conversion, or for some representations we can write our own parser?

Also, I just want to confirm whether we should do all of this in a single PR or break it down into multiple smaller PRs?

ronaldtse:
> We should first create all the parsers to parse the input into a `Plurimath::Formula` and then in the next step we should convert the `Plurimath::Formula` into different representations.

Yes.

> We can use gems if exists, like [asciimath](https://github.com/asciidoctor/asciimath) for the conversion, or for some representations we can write our own parser?

We can always use gems if they exist. However, there will come a point that parsing is easier than integrating.

For example, the `asciimath` gem converts AsciiMath to MathML. If we use this gem, we will have to parse MathML into a Plurimath::Formula anyway.

> Also, I just want to confirm whether we should do all of this in a single PR or break it down into multiple smaller PRs?

Let's break it down into multiple issues and PRs. Let's just merge whatever we can when we can.

lib/plurimath/converters/omml.rb
ronaldtse: You probably want to update Omml2Mathml code to create this consistency. Feel free to change it!

lib/plurimath/asciimath_parser/sin.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module AsciimathParser
+    # Sin Class
+    class Sin
```
HassanAkbar, line 6: @suleman-uzair `Sin`, `Cos`, `Tan` are functions. 
So, I think they should be inside `Plurimath::AsciimathParser::Functions` ?

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.6', '2.5', '2.4' ]
+        ruby: [ '2.6', '3.0' ]
```
ronaldtse, line 21: Add 2.7 please.

lib/plurimath/asciimath_parser/sin.rb
ronaldtse: Agree.

lib/plurimath/asciimath_parser/cos.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module AsciimathParser
```
ronaldtse, line 4: Maybe just `Asciimath` instead of `AsciimathParser`?

I'm thinking you may have this module hierarchy:

```ruby
Plurimath::Math # for math representation (this is the "central model" of math)
Plurimath::Math::Formula # one math formula
Plurimath::Asciimath # the AsciiMath parser
Plurimath::Latexmath # the LaTeX math parser
Plurimath::Omml # the OMML parser
Plurimath::Mathml # the MathML parser
```

And then you have the objects per math language:
```ruby
Plurimath::Asciimath::Function::Cos
Plurimath::Asciimath::Symbol
...
```

ronaldtse (changes requested):

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.6', '2.5', '2.4' ]
+        ruby: [ '2.6', '2.7', '3.0' ]
```
HassanAkbar, line 21: Remove `2.7` from the include section below.

lib/plurimath/asciimath_parser/cos.rb
HassanAkbar: @ronaldtse @suleman-uzair I was thinking that we might not need the functions for mathematical representation. So, we can move them to the common Math namespace. For example, we can have:

```ruby
Plurimath::Math::Function::Cos
```

because these functions are related to central model and we don’t need them for every representation.
Let me know if I am missing something here?

ronaldtse: @HassanAkbar indeed. I meant `Plurimath::Asciimath::Function::Cos` as a parse node of the Asciimath language "node" of representing `Plurimath::Math::Function::Cos`. If we don't need it, fine.

HassanAkbar (approved):

PR #7, opened 2022-03-11
Asciimath parser

### LINKED ISSUE #10: Parse asciimath string to plurimath formula.
Parse `Asciimath` string into ruby Formula

suleman-uzair:
Added **`Plurimath::Asciimath::Parser`** class which converts `Asciimath` string into ruby Formula.

closes #10


ronaldtse:
@suleman-uzair can you use the Parslet gem for parsing? The parsing rules will be a lot more clear.

spec/plurimath/asciimath/parser_spec.rb
```diff
@@ -0,0 +1,505 @@
...
+  it "returns instance of Cos formula against the string" do
+    asciimath = init_spec_resp("cos(2)")
+    # initializing common used values in respective named variables
+    cos = asciimath.value.first
+    cos_value = cos.angle.value
+    # cos formula
+    expect(cos.class).to eql(Plurimath::Math::Function::Cos)
+    expect(cos_value[0].value).to eq("(")
+    expect(cos_value[1].value).to eq("2")
+    expect(cos_value[2].value).to eq(")")
```
mico, line 15: Probably we don't need to store parentheses here, @ronaldtse right?

mico:
> > @suleman-uzair can you use the Parslet gem for parsing? The parsing rules will be a lot more clear.
> 
> @ronaldtse I’ve pushed the new code using the **`Parslet`** gem as you mentioned above. The work is still in progress. Just needed to confirm that I’m using it right way. Let me know if you have any feedback.

@suleman-uzair parser's code looks ok, only hard to read

spec/plurimath/asciimath/parser_spec.rb
ronaldtse: Agree. I don't think a parenthesis is just a symbol. The point of a parenthesis is to wrap an expression. An expression can contain the parenthesis symbols as attributes, but the semantics are different.

ronaldtse: Here, cosine is a function that includes an expression. An expression could contain numbers. The current parse tree does not reflect this at all.

mico:
With the current implementation it's hard to write readable, efficient tests.
If I would write them I would do something like:
```
RSpec.describe Plurimath::Asciimath::Parser do
  subject { described_class.parse(formula) }
  
  context "returns instance of Cos formula against the string" do
    let(:formula) { "cos(2)" }

    it "returns Cos formula" do
      expect(subject.to_h).to eq({ formulas: Plurimath::Math::Function::Cos.new(2) })
    end
  end
end
```

I'm not sure about last part with subject.to_h and comparing with hash, because I don't know project and goals so well. But idea to find the way how to make tests more readable and compact.

suleman-uzair:
> With the current implementation it's hard to write readable, efficient tests. If I would write them I would do something like:
> 
> ```
> RSpec.describe Plurimath::Asciimath::Parser do
>   subject { described_class.parse(formula) }
>   
>   context "when contains cosine function" do
>     let(:formula) { "cos(2)" }
> 
>     it "returns Cos formula" do
>       expect(subject.to_h).to eq({ formulas: Plurimath::Math::Function::Cos.new(2) })
>     end
>   end
> end
> ```
> 
> I'm not sure about last part with subject.to_h and comparing with hash, because I don't know project and goals so well. But idea to find the way how to make tests more readable and compact.

Great suggestion. I think this will make many test cases, if not all much more readable.

suleman-uzair:
> > > @suleman-uzair can you use the Parslet gem for parsing? The parsing rules will be a lot more clear.
> > 
> > 
> > @ronaldtse I’ve pushed the new code using the **`Parslet`** gem as you mentioned above. The work is still in progress. Just needed to confirm that I’m using it right way. Let me know if you have any feedback.
> 
> @suleman-uzair parser's code looks ok, only hard to read

I’m working on making it more readable. This was just to confirm that I am going in the right direction.

lib/plurimath/math/function/unary_function.rb
```diff
@@ -0,0 +1,17 @@
...
+module Plurimath
+  module Math
+    module Function
+      class UnaryFunction
+        def ==(object)
+          object.parameter_value == parameter_value
+        end
+
+        def parameter_value
+          instance_variable_get(instance_variables[0])
```
ronaldtse, line 12: @suleman-uzair is `instance_variables` the default Ruby `Object.instance_variables`? This seems to be a strange usage of it that can be problematic.

suleman-uzair: > @suleman-uzair is `instance_variables` the default Ruby `Object.instance_variables`? This seems to be a strange usage of it that can be problematic.

Yes, you are right, it can be done in a better way. I’ve changed it and updated the code.

lib/plurimath/asciimath/constants.rb
```diff
@@ -0,0 +1,315 @@
...
+        \]
+        :}
+      ].freeze
+      LPAREN = %i[
+        (
+        {
+        (:
+        \[
+        {:
+      ].freeze
```
HassanAkbar, line 311: Can we use `PARENTHESIS.keys` and `PARENTHESIS.values` instead of `LPAREN` and `RPAREN`?

lib/plurimath/asciimath/transform.rb
```diff
@@ -0,0 +1,294 @@
...
+
+      rule(power_base: sequence(:power_base), expr: simple(:expr)) do
+        power_base + [expr]
+      end
+
+      rule(power_base: sequence(:power_base), expr: sequence(:expr)) do
+        power_base + expr
+      end
+
+      rule(sequance: simple(:sequance), expr: sequence(:expr)) do
```
ronaldtse, line 64: `sequence` instead of `sequance`.

ronaldtse (approved):
One last comment!

ronaldtse:
@suleman-uzair before merging, please rebase and clean up the commit history. Thanks.

HassanAkbar (approved):

ronaldtse:
@suleman-uzair next time can you please rebase and merge the commits that do similar things for cleanliness?

Screenshot 2022-04-05 at 8 23 35 PM

In addition, we want to get rid of these "Merges" that pollute Git history.

Screenshot 2022-04-05 at 8 24 41 PM

Thanks.

PR #8, opened 2022-03-11
Formula to asciimath

### LINKED ISSUE #9: Parse plurimath formula to asciimath string.
Convert Formula object into `Asciimath` string

suleman-uzair:
Added **`to_asciimath`** method in all `function` files to convert ruby formula back to `Asciimath` string.

closes #9

### DIFF
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -12,6 +12,10 @@ module Plurimath
       def ==(object)
         object.value == value
       end
+
+      def to_asciimath
+        value.map(&:to_asciimath).join("")
+      end
     end
   end
 end
--- a/lib/plurimath/math/function/binary_function.rb
+++ b/lib/plurimath/math/function/binary_function.rb
@@ -11,6 +11,16 @@ module Plurimath
           @parameter_two = parameter_two
         end
 
+        def to_asciimath
+          "#{self.class.name.split('::').last.downcase}#{value_to_asciimath}"
+        end
+
+        def value_to_asciimath
+          first_value = "(#{parameter_one.to_asciimath})" if parameter_one
+          second_value = "(#{parameter_two.to_asciimath})" if parameter_two
+          "#{first_value}#{second_value}"
+        end
+
         def ==(object)
           object.parameter_one == parameter_one &&
             object.parameter_two == parameter_two
--- a/spec/plurimath/math/formula_spec.rb
+++ b/spec/plurimath/math/formula_spec.rb
@@ -2,13 +2,749 @@ require_relative '../../../lib/plurimath/math'
 
 RSpec.describe Plurimath::Math::Formula do
 
-  it 'returns instance of Formula' do
-    formula = Plurimath::Math::Formula.new('1 + 2')
-    expect(formula).to be_a(Plurimath::Math::Formula)
-  end
+  describe ".to_asciimath" do
+    subject(:formula) { Plurimath::Asciimath::Parser.new(exp).parse.to_asciimath }
 
-  it 'returns Formula instance' do
-    formula = Plurimath::Math::Formula.new('1 + 2')
-    expect(formula.value).to eql(['1 + 2'])
+    context "contains string" do
+      it 'returns Formula object' do
+        formula = Plurimath::Math::Formula.new('1 + 2')
+        expect(formula).to be_a(Plurimath::Math::Formula)
+      end
+    end
+
+    context "contains string" do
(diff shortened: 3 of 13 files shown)


lib/plurimath/math/formula.rb
```diff
@@ -12,6 +12,14 @@ def initialize(value = [])
       def ==(object)
         object == value
       end
+
+      def to_asciimath
+        entities = ""
+        value.each do |entity|
+          entities += entity.to_asciimath
```
ronaldtse, line 19: This seems like just a joining of strings, so this would be cleaner:

```ruby
def to_asciimath
  value.map(&:to_asciimath).join("")
end
```

lib/plurimath/math/function/g.rb
```diff
@@ -10,6 +10,10 @@ def initialize(value)
           @value = value
         end
 
+        def to_asciimath
+          "g#{value&.to_asciimath}"
```
ronaldtse, line 14: Seems that there are many instances of `class Function` subclasses that has this method. Can we use a method in Function like this:
```ruby
def to_asciimath
  "#{function_name_in_asciimath}#{value.to_asciimath}"
end
```

ronaldtse (changes requested):

lib/plurimath/math/function/binary_function.rb
```diff
@@ -11,6 +11,16 @@ def initialize(parameter_one = nil, parameter_two = nil)
           @parameter_two = parameter_two
         end
 
+        def to_asciimath
+          "#{self.class.name.split('::').last.downcase}#{value_to_asciimath}"
+        end
```
HassanAkbar, line 16: It seems to me that this method is currently not being used anywhere because of the method overriding in each class that inherits from `BinaryFunction`.

```ruby
first_value = "_(#{parameter_one.to_asciimath})" if parameter_one
second_value = "^(#{parameter_two.to_asciimath})" if parameter_two
"int#{first_value}#{second_value}"
```

It looks like the above code is repeated many times in binary classes. Can we move that here and only add the `to_asciimath` method where the pattern is different.

suleman-uzair: > It seems to me that this method is currently not being used anywhere because of the method overriding in each class that inherits from `BinaryFunction`.
> 
> ```ruby
> first_value = "_(#{parameter_one.to_asciimath})" if parameter_one
> second_value = "^(#{parameter_two.to_asciimath})" if parameter_two
> "int#{first_value}#{second_value}"
> ```
> 
> It looks like the above code is repeated many times in binary classes. Can we move that here and only add the `to_asciimath` method where the pattern is different.

The ``to_asciimath`` method in **``BinaryFunction``** class is being used more than the methods defined in each class.
It's used in other files which had no changes and are not included in this PR.

HassanAkbar (approved):

HassanAkbar (approved):
LGTM

spec/plurimath/math/formula_spec.rb
```diff
@@ -2,13 +2,749 @@
 
 RSpec.describe Plurimath::Math::Formula do
 
-  it 'returns instance of Formula' do
-    formula = Plurimath::Math::Formula.new('1 + 2')
-    expect(formula).to be_a(Plurimath::Math::Formula)
-  end
+  describe ".to_asciimath" do
+    subject(:formula) { Plurimath::Asciimath::Parser.new(string).parse.to_asciimath }
```
ronaldtse, line 6: Prefer using `exp` (for expression) rather than `string`.

spec/plurimath/math/formula_spec.rb
```diff
@@ -2,13 +2,749 @@
...
+
+    context "contains string" do
+      it 'returns Formula object' do
+        formula = Plurimath::Math::Formula.new('1 + 2')
+        expect(formula.value).to eql(['1 + 2'])
+      end
+    end
+
+    context "contains cos formula string" do
+      let(:string) { "cos(2)" }
```
ronaldtse, line 23: Prefer using `exp` (for expression) rather than `string`.

ronaldtse (changes requested):

ronaldtse:
Ping @suleman-uzair can we merge this soon?

suleman-uzair:
> Ping @suleman-uzair can we merge this soon?

I’ve updated the code as you suggested.

ronaldtse:
@suleman-uzair can you help rebase and merge? Thanks.

PR #13, opened 2022-04-22
added parslet parse rules for mathml string

### LINKED ISSUE #14: Add MathML parser
Add parser for parsing **```MathML```** string to the ```syntax tree```.

suleman-uzair:
Added **```Parslet```** parse rules for parsing **```MathMl```** string

closes #14

### DIFF
--- a/lib/plurimath/mathml.rb
+++ b/lib/plurimath/mathml.rb
@@ -9,7 +9,7 @@ module Plurimath
     end
 
     def to_formula
-      # TODO: Will be implemented soon
+      Plurimath::Mathml::Parser.new(text).parse
     end
   end
 end
--- a/lib/plurimath/mathml.rb
+++ b/lib/plurimath/mathml.rb
@@ -9,7 +9,7 @@ module Plurimath
     end
 
     def to_formula
-      Plurimath::Mathml::Parser.new(text).parse
+      # TODO: Will be implemented soon
     end
   end
 end
--- a/lib/plurimath/mathml/parse.rb
+++ b/lib/plurimath/mathml/parse.rb
@@ -5,9 +5,9 @@ module Plurimath
   class Mathml
     class Parse < Parslet::Parser
       rule(:parse_record) do
-        array_to_expression(Constants::CLASSES, String).as(:class) |
-          array_to_expression(Constants::UNICODE_SYMBOLS.keys, Symbol).as(:symbol) |
-          array_to_expression(Constants::SYMBOLS.keys, Symbol).as(:symbol) |
+        array_to_expression(Constants::CLASSES).as(:class) |
+          array_to_expression(Constants::UNICODE_SYMBOLS.keys).as(:symbol) |
+          array_to_expression(Constants::SYMBOLS.keys).as(:symbol) |
           match["a-zA-Z"].repeat(1).as(:text) |
           match(/[0-9]/).repeat(1).as(:number) |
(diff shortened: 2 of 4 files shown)


lib/plurimath/mathml.rb
```diff
@@ -9,7 +9,7 @@ def initialize(text)
     end
 
     def to_formula
-      # TODO: Will be implemented soon
+      Plurimath::Mathml::Parser.new(text).parse
```
HassanAkbar, line 12: Will this work currently?

lib/plurimath/mathml/parse.rb
```diff
@@ -0,0 +1,77 @@
...
+          expr | str(tag)
+        end
+      end
+
+      def parse_symbols
+        Constants::SYMBOLS.keys.reduce do |expr, tag|
+          expr = str(expr) if expr.is_a?(Symbol)
+          expr | str(tag)
+        end
+      end
```
HassanAkbar, line 45: These functions and the `tags` function on line 65 have similar code can we merge these into a common method?

HassanAkbar (changes requested):
Just some minor comments. Other than that it looks good. 👍

lib/plurimath/mathml/parse.rb
ronaldtse: Agree

spec/plurimath/mathml/parse_spec.rb
```diff
@@ -0,0 +1,166 @@
...
+                <mi>x</mi>
+                <mi>s</mi>
+              </munderover>
+            </mrow>
+          </mstyle>
+        </math>
+      MATHML
+    }
+    it "returns formula of sum and prod" do
+      expected_value = {
```
ronaldtse, line 134: Is it possible to make this comparison more simple? For example, if the structure changes in the future, these compares will be very tough to update. You should be testing for particular outcomes, not data equivalence of the structures.

ronaldtse (changes requested):

spec/plurimath/mathml/parse_spec.rb
ronaldtse: But we do do that in a future PR, too.

spec/plurimath/mathml/parse_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    let (:exp) { "<math xmlns='http://www.w3.org/1998/Math/MathML'></math>" }
+
+    it "matches open and close tag" do
+      expect(formula[:open]).to eq("math")
+      expect(formula[:close]).to eq("math")
+    end
+
+    it "matches attributes" do
+      expect(formula[:attributes].first[:name]).to eq("xmlns")
+      expect(formula[:attributes].last[:value]).to eq("http://www.w3.org/1998/Math/MathML")
```
HassanAkbar, line 19: I think both should be consistent. Currently, it feels like the name and value are for different attributes.

```suggestion
      expect(formula[:attributes].first[:name]).to eq("xmlns")
      expect(formula[:attributes].first[:value]).to eq("http://www.w3.org/1998/Math/MathML")
```

HassanAkbar (approved):
LGTM. 👍

ronaldtse (approved):

ronaldtse:
Thanks @suleman-uzair !

PR #16, opened 2022-04-22
Added parslet transformation rules for syntax tree

suleman-uzair:


suleman-uzair:
This PR's code is transferred to #17.

PR #17, opened 2022-04-22
Added transform rules for syntax tree

### LINKED ISSUE #15: Transform parsed syntax tree to Plurimath Formula
Transform parsed ```syntax tree``` to **```Plurimath Formula```** object.

suleman-uzair:
Added transformation rules for **``syntax tree``**.

closes #15

### DIFF
--- a/lib/plurimath/math/function/underover.rb
+++ b/lib/plurimath/math/function/underover.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module Math
+    module Function
+      class Underover
+        attr_accessor :parameter_one, :parameter_two, :parameter_three
+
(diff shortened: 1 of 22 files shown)


lib/plurimath/math/function/underover.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module Math
+    module Function
+      class Underover
```
ronaldtse, line 6: The code for `Underover` and `PowerBase` is identical. Can we DRY this with either a common superclass or using one as a superclass/subclass?

ronaldtse (changes requested):

ronaldtse (approved):
Thanks @suleman-uzair !

PR #47, opened 2022-09-12
Implemented OMML simple and sub sup parser rules

### LINKED ISSUE #48: Implement OMML parser rules for simple and Sub Sup tags
Implement Parslet parser rules for OMML simple and Sub Sup tags.

suleman-uzair:
Added Parslet parser rules for Simple and Sub/Sup tags.

closes #48


HassanAkbar (approved):

lib/plurimath/omml/parse.rb
```diff
@@ -0,0 +1,87 @@
...
+        str(string).as(name)
+      end
+
+      def quoted_string
+        (str('"') >> match("[^\"]").repeat.as(:value) >> str('"')) |
+          (str("'") >> match("[^\']").repeat.as(:value) >> str("'"))
+      end
+
+      def omission_tag
+        tag = str("<w:") | str("<m:")
```
ronaldtse, line 72: The direct usage of `<w:` and `<m:` for parsing is very dangerous, because these are actually XML namespace aliases. The actual definitions of the xmlns (XML namespace) are like this:
https://github.com/transpect/docx2hub/blob/519ad7c22869d5b1536809ab713ac64951cff28d/xsl/omml2mml/omml2mml.xsl#L12-L14

i.e
```xml
  xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" 
  xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math" 
```

For example, if the xmlns line looked like this:
```
  xmlns:moo="http://schemas.openxmlformats.org/wordprocessingml/2006/main" 
```

Then the tags will begin like `<moo:oMath ...`.

I would strongly suggest that we use an XML parser like Ox (the native Ruby XML parser, over Nokogiri), to parse this XML.

ronaldtse (changes requested):

ronaldtse:
@suleman-uzair I don't mean to request a rework -- but having a parser that is native to the format will enable better maintainability for the future. Hopefully this will allow you to do "more effective but less work" in the future!

suleman-uzair:
Hi @ronaldtse.

1. Should we use **Ox** library which creates nodes(just like **Nokogiri** as you mentioned), but we will have to parse it without using ```Parslet```.
2. Or should we use **``` OMML2Mathml ```** library(or it’s code) to parse **``` OMML to Mathml ```** and then from **``` Mathml to Plurimath::Formula ```**.

What do you suggest?

ronaldtse:
@suleman-uzair the preferred approach is 1, because we don't want to reimplement an XML parser and support for its eccentric features like namespaces. Moreover, we still do the reverse conversion of MathML to OMML, which means reusing the OMML2MathML code will a challenge.

The main issue here is a data model mapping between OMML and Plurimath::Formula, so if we have done that conversion, we can use Plurimath::Formula as the central conversion function between OMML and MathML (OMML => Plurimath => MathML).

lib/plurimath/omml/parse.rb
```diff
@@ -0,0 +1,134 @@
...
+        elsif ["sSubSup", "sPre"].include?(tag_name)
+          Math::Function::PowerBase
+        end
+      end
+
+      def tag_name(node)
+        if node.is_a?(String)
+          node
+        else
+          node&.value&.gsub(/[a-zA-Z]{1,}:/, "")
```
HassanAkbar, line 129: There is a `strip_namespace` option in `OX`. Can we use that here?

ronaldtse (approved):
Thank you @suleman-uzair !

PR #66, opened 2022-10-12
Updated mathml parser rule and added specs for decimal numeric values

### LINKED ISSUE #64: Mathml `mn` is not accepting decimals
Unlike the preceding issue reported, this is a dealbreaker.

`mn` is only accepting sequences of decimals, it is not allowing comma or dot. (Per https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mn, it must accept both).

````
Plurimath::Math.parse("<math><mn>6,4</mn></math>", "mathml") # crashes
Plurimath::Math.parse("<math><mn>6.4</mn></math>", "mathml") # crashes
Plurimath::Math.parse("<math><mn>64</mn></math>", "mathml") # OK
````

### PR DESCRIPTION
**Pull request by suleman-uzair**
Updated **MathML** parser rule and added spec for decimal numbers(separated by comma and dot).

closes #64

### DIFF
--- a/spec/plurimath/mathml/parser_spec.rb
+++ b/spec/plurimath/mathml/parser_spec.rb
@@ -1089,6 +1089,45 @@ RSpec.describe Plurimath::Mathml::Parser do
     end
   end
 
+  context "contains mathml string of dot decimal values only" do
+    let(:exp) {
+      <<~MATHML
+        <math xmlns='http://www.w3.org/1998/Math/MathML'>
+          <msgroup>
+            <mn>1.4</mn>
+            <mn>1.4.5</mn>
+          </msgroup>
+        </math>
+      MATHML
+    }
+    it "returns formula of decimal values" do
+      expected_value = Plurimath::Math::Formula.new([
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spec/plurimath/mathml/parser_spec.rb [hunk 1]
```diff
@@ -1089,6 +1089,45 @@
     end
   end
 
+  context "contains mathml string of dot decimal values only" do
+    let(:exp) {
+      <<~MATHML
+        <math xmlns='http://www.w3.org/1998/Math/MathML'>
+          <msgroup>
+            <mn>1.4</mn>
+            <mn>1.4.5</mn>
```
ronaldtse, line 1098: `1.4.5` is not a number, it should not be accepted.

Please refer to https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mn for the accepted values.

ronaldtse:
As per #64 , please directly refer to https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mn for what values are accepted. There should be a "number" class that reflects the proper values.

PR #70, opened 2022-10-17
Updated parser rule and added spec for decimal values

### LINKED ISSUE #69: AsciiMath should accept decimal numbers using the decimal point
````
Plurimath::Math.parse("0.1", "asciimath")
Plurimath::Math.parse("0,1", "asciimath")
Plurimath::Math.parse(".1", "asciimath")
Plurimath::Math.parse(",1", "asciimath")
Plurimath::Math.parse("-0.1", "asciimath")
Plurimath::Math.parse("-0,1", "asciimath")
Plurimath::Math.parse("-.1", "asciimath")
Plurimath::Math.parse("-,1", "asciimath")
````

All of these are crashing on the parser.

You should be studying the spec of https://github.com/asciidoctor/asciimath/blob/master/spec/parser_spec.rb . But that spec does not even include numbers...

### PR DESCRIPTION
**Pull request by suleman-uzair**
Updated **AsciiMath** parser number rule, and constants and added specs for decimal values.

closes #69

### CODE REVIEW

**Review by HassanAkbar — APPROVED** [review]

File: spec/plurimath/asciimath/parser_spec.rb [hunk 1]
```diff
@@ -1141,5 +1141,103 @@
...
+    end
+
+    context "when contains comma sepearted numeric value" do
+      let(:string) { "0,1" }
+
+      it "returns formula" do
+        expected_value = Plurimath::Math::Formula.new([
+          Plurimath::Math::Number.new("0"),
+          Plurimath::Math::Symbol.new(","),
+          Plurimath::Math::Number.new("1"),
```
ronaldtse, line 1163: Actually, the original request is to parse these values all identically, i.e.:
* `0,1` == `0.1`
* `-.1` == `-0.1`
* `-0,1` == `-0.1`
* `-.1` == `-0.1`

Therefore the current expectations all need to be updated accordingly.

ronaldtse (changes requested):

ronaldtse:
@suleman-uzair this PR should really only handle `0.1` (i.e. decimal point) not `0,1` the decimal comma.

The decimal comma requires further thought on what to use as number separators, I am not convinced that we have an immediate solution right now. Why don't we just skip this for now and deal with other urgent issues first?

Again, the "decimal comma" is NOT important.

ronaldtse:
@suleman-uzair is this to be merged?

suleman-uzair:
> @suleman-uzair is this to be merged?

@ronaldtse Yes, it's ready to merge.

spec/plurimath/asciimath/parser_spec.rb
```diff
@@ -1141,5 +1141,103 @@
...
+      end
+    end
+
+    context "when contains nagative decimal value" do
+      let(:string) { "-0.1" }
+
+      it "returns formula" do
+        expected_value = Plurimath::Math::Formula.new([
+          Plurimath::Math::Symbol.new("-"),
+          Plurimath::Math::Number.new("0.1"),
```
ronaldtse, line 1198: @suleman-uzair shouldn't this be:
```ruby
expected_value = Plurimath::Math::Formula.new([
  Plurimath::Math::Number.new("-0.1"),
])
```

spec/plurimath/asciimath/parser_spec.rb
```diff
@@ -1141,5 +1141,103 @@
...
+    context "when contains nagative comma seperated value" do
+      let(:string) { "-0,1" }
+
+      it "returns formula" do
+        expected_value = Plurimath::Math::Formula.new([
+          Plurimath::Math::Symbol.new("-"),
+          Plurimath::Math::Number.new("0"),
+          Plurimath::Math::Symbol.new(","),
+          Plurimath::Math::Number.new("1"),
+        ])
```
ronaldtse, line 1213: Shouldn't this be:
```ruby
expected_value = Plurimath::Math::Formula.new([
  Plurimath::Math::Number.new("-0"),
  Plurimath::Math::Symbol.new(","),
  Plurimath::Math::Number.new("1"),
])
```

suleman-uzair:
@ronaldtse

ronaldtse:
@suleman-uzair  I’m not sure what the impact of that change is. Can you elaborate?

suleman-uzair:
> @suleman-uzair I’m not sure what the impact of that change is. Can you elaborate?

Changes in the parsing behavior of operators(- and +) in Asciimath won’t affect on output(but only in parser rules and spec values) because we are not making any changes to operators on output, but I have some questions regarding parsing rules,
should 1-2 be parsed as 1 and - and 2 OR 1 and -2 ?
should -a be parsed as -a ?
should a-beta be parsed as a and - and beta OR a and -beta ?

ronaldtse:
Thank you for the explanation @suleman-uzair ! I suppose the issue with `-` depends on whether it is a minus or unary minus operator. Parsing `-` as a an operator is indeed correct. Let's merge this. Thanks!

PR #92, opened 2022-11-17
[WIP] Updates/formula to mathml ox nodes

suleman-uzair:
**[Work In Progress]**

Updated `to_mathml` methods using `Ox::Element` objects for `Formula` to `MathML`.

### DIFF
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -23,21 +23,23 @@ module Plurimath
       end
 
       def to_mathml
-        <<~MATHML
-          <math xmlns='http://www.w3.org/1998/Math/MathML' display='block'>
-            <mstyle displaystyle='true'>
-              #{mathml_content}
-            </mstyle>
-          </math>
-        MATHML
+        math  = Utility.omml_element("math", attributes: { xmlns: 'http://www.w3.org/1998/Math/MathML', display: 'block' })
+        style = Utility.omml_element("mstyle", attributes: { displaystyle: 'true' })
+        Utility.update_nodes(style, mathml_content)
+        Utility.update_nodes(math, [style])
+        Ox.dump(math, indent: 2)
+          .gsub("&amp;", "&")
       end
 
       def to_mathml_without_math_tag
-        "<mrow>#{mathml_content}</mrow>"
+        Utility.update_nodes(
+          Utility.omml_element("mrow"),
+          mathml_content,
+        )
       end
 
       def mathml_content
-        value.map(&:to_mathml_without_math_tag).join
+        value.map(&:to_mathml_without_math_tag)
       end
 
       def to_latex
--- a/lib/plurimath/math/function/base.rb
+++ b/lib/plurimath/math/function/base.rb
@@ -13,9 +13,16 @@ module Plurimath
         end
 
         def to_mathml_without_math_tag
-          first_value = parameter_one.to_mathml_without_math_tag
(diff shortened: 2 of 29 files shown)


suleman-uzair:
Asciimath and latex don’t have alternatives to the unicode characters.
So in conversion, we are converting supported unicodes to symbols (supported words like sum, prod, etc) but the unsupported can be seen as text in the screenshot below.

Screenshot from 2022-11-21 19-45-53

@opoudjis  In the examples provided by you, I can see Unicode characters.

@ronaldtse  should we plan to cater to these Unicodes ?
I was thinking, we can use `htmlentities` gem as the solution, it will convert unicodes to their presentable state(like ∑, ∏, etc..).
@ronaldtse  what do you think ?

suleman-uzair:
> @suleman-uzair sorry I'm not sure what you mean here... can you give some concrete examples?
> 
> Are you saying that the math symbols across the different syntaxes, i.e. MathML, AsciiMath and LaTeX math, are not aligned (some symbols only available in some syntaxes)?

Yes you got it right @ronaldtse .
For example, the symbol:
Screenshot from 2022-11-22 12-34-51
1. is available in Latex `barleftarrowrightarrowba`.
2. is available in MathML as `&#x21b9;`.
3. is not available in AsciiMath.

Similarly, I was looking at examples provided by @opoudjis .
Screenshot from 2022-11-22 13-04-15
In some of them, MathML uses `&#xB0;C` where `&#xB0;` is not supported.

ronaldtse:
@suleman-uzair this is indeed a problem that needs to be handled.

Can you develop a table that maps all symbols across the 4 syntaxes (MathML, OOML, AsciiMath, LaTeX math), so we know what's missing? This is also very useful for documentation.

For symbols missing in AsciiMath, the best we can do is to support them by using the LaTeX math command name.

suleman-uzair:
> @suleman-uzair this is indeed a problem that needs to be handled.
> 
> Can you develop a table that maps all symbols across the 4 syntaxes (MathML, OOML, AsciiMath, LaTeX math), so we know what's missing? This is also very useful for documentation.
> 
> For symbols missing in AsciiMath, the best we can do is to support them by using the LaTeX math command name.

MathML and OMML supports all unicode characters and not commands like AsciiMath and Latex.
The list of Ascii symbols is quite short compared to Unicodes.
We can create a list of symbols not supported from Latex to AsciiMath but the list of unicodes is quite large to mention.
So I’m unable to understand what values should we add to the table you asked for!

Also @ronaldtse , you said that
` For symbols missing in AsciiMath, the best we can do is to support them by using the LaTeX math command name. `
From what I understand is that we can use LaTeXMath’s symbols(name commands like \barleftarrowrightarrowba, etc) in AsciiMath. Is this correct or do you mean something different?

ronaldtse:
Thank you @suleman-uzair !

ronaldtse (approved):

ronaldtse:
@suleman-uzair we will need documentation on these new AsciiMath symbols because they are not used by the original AsciiMath parsers (asciimath.js and AsciiMath in Ruby).

suleman-uzair:
> @suleman-uzair we will need documentation on these new AsciiMath symbols because they are not used by the original AsciiMath parsers (asciimath.js and AsciiMath in Ruby).

@ronaldtse I have opened an issue for documentation update and will work on that after the bug fixes.

ronaldtse:
Great, thanks!

PR #93, opened 2022-11-21
Updated Latex parser rules and added specs

### LINKED ISSUE #61: Add additional tests for LaTeX math parsing/generation to MathML from Metanorma
Metanorma contains a test for LaTeX to MathML conversion:

* https://github.com/metanorma/metanorma-standoc/blob/ec29eaf5eea7384265a6ef2ba557901aeb6afd03/spec/metanorma/blocks_spec.rb#L86-L242

This should be added to the LaTeX math parsing  (1 test) and MathML generation test suite (another test).

### LINKED ISSUE #82: Rendering `\sqrt{..}^{..}` correctly (carried from latexmath)
Reported by @infinity0 (thank you!) at https://github.com/plurimath/latexmath/issues/43 :

latexmath gets confused by the combination of \sqrt and ^. With `\sqrt{a}^{b}` instead the output looks like `\sqrt(^) a b`:

```xml
<mrow><msqrt><mi>^</mi></msqrt><mrow><mi>&#x00061;</mi></mrow><mrow><mi>&#x00062;</mi></mrow></mrow>
```

It should instead be:

```xml
  <msup>
    <msqrt>
      <mi>a</mi>
    </msqrt>
    <mi>b</mi>
  </msup>
```

This issue happens at the latexmath gem, but it may or may not also occur at plurimath, which is the successor of latexmath.

We need to:
1. Verify if this issue occurs at plurimath
2. Ensure that this particular case is tested.

### LINKED ISSUE #84: regression from latexmath, crashes on normal latex input
plurimath can't parse the following, but latexmath can:

```ruby
require 'plurimath'
require 'latexmath'

latex = "$$   (\mathcal{F}f)(y)
    = \frac{1}{\sqrt{2\pi}^{\ n}}
      \int_{\mathbb{R}^n} f(x)\,
      e^{-\mathrm{i} y \cdot x} \,\mathrm{d} x.
$$"

puts Latexmath.parse(latex).to_mathml
puts Plurimath::Math.parse(latex, "latex").to_mathml
```

~~~~
$ ruby test.rb 
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block"><mrow> [.. etc, output from latexmath is fine ..]
~/.local/share/gem/ruby/3.0.0/gems/parslet-2.0.0/lib/parslet/cause.rb:70:in `raise': Expected one of [ITERATION EXPRESSION, ITERATION] at line 1 char 1. (Parslet::ParseFailed)
	from ~/.local/share/gem/ruby/3.0.0/gems/parslet-2.0.0/lib/parslet/atoms/base.rb:49:in `parse'
	from ~/.local/share/gem/ruby/3.0.0/gems/plurimath-0.2.1/lib/plurimath/latex/parser.rb:16:in `parse'
	from ~/.local/share/gem/ruby/3.0.0/gems/plurimath-0.2.1/lib/plurimath/latex.rb:12:in `to_formula'
	from ~/.local/share/gem/ruby/3.0.0/gems/plurimath-0.2.1/lib/plurimath/math.rb:39:in `parse'
	from test.rb:11:in `<main>'
exit code 1
~~~~

### LINKED ISSUE #85: LaTeX math: verify support of horizontal spacing commands
As reported here: https://github.com/plurimath/latexmath/issues/33

From @manuelfuenmayor :

Currently, commands like `quad` and `qquad` are not supported by latexmath:
````
[latexmath%unnumbered]
++++
f_i =  \sum_{j=1}^2 s_{ij} n_j \quad {\rm for} \quad i = 1,2
++++
````
renders
issue4

Also not supported:
* negative space: `\!`
* boldface text: `\textbf`

From @infinity0:

Also not supported:

* explicit space (backslash-spac) `\ ` - currently latexmath generates an actual backslash character instead

(python-docutils has the correct behaviour)

### PR DESCRIPTION
**Pull request by suleman-uzair**
Added and updated parser rules and specs for Latex.

closes #85
closes #84
closes #82
closes #61

### CODE REVIEW

File: lib/plurimath/latex/parse.rb [hunk 1]
```diff
@@ -92,7 +92,8 @@ class Parse < Parslet::Parser
       end
 
       rule(:symbol_class_commands) do
-        (slash >> symbols) |
+        (str("&#x") >> match["0-9a-zA-Z"].repeat >> str(";")).as(:unicode_symbols) |
```
HassanAkbar, line 95: I think it should be `match["0-9a-fA-F"]` for hexadecimal numbers.

lib/plurimath/math/unicode.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module Math
+    class Unicode < Symbol
+
+      def ==(object)
+        object.value == value
+      end
```
HassanAkbar, line 9: This will be inherited from the parent method so no need to redefine it here.

lib/plurimath/latex/parse.rb
```diff
@@ -92,8 +92,9 @@ class Parse < Parslet::Parser
       end
 
       rule(:symbol_class_commands) do
-        (slash >> symbols) |
-          (unary >> intermediate_exp.as(:first_value)).as(:unary_functions) |
+        (unary >> intermediate_exp.as(:first_value)).as(:unary_functions) |
+          (str("&#x") >> match["0-9a-fA-F"].repeat >> str(";")).as(:unicode_symbols) |
```
opoudjis, line 96: Metanorma will only generate hex unicode escapes, but if this gem is to be used in general, it should deal with decimal unicode escapes as well: you can't ban them.

lib/plurimath/latex/parser.rb
```diff
@@ -9,7 +9,15 @@ class Parser
       attr_accessor :text
 
       def initialize(text)
-        @text = text.gsub(/\s/, "")
+        enti = HTMLEntities.new
+        text = enti.encode(text, :hexadecimal)
```
opoudjis, line 13: I've found it's useful to do `enti.encode(enti.decode(text), :hexadecimal)` : you may well get entity encoding in your input.

opoudjis (approved):

PR #100, opened 2022-12-22
Fixed/implemented metanorma examples

suleman-uzair:
Extracted examples from metanorma repos(metioned below) and added/updated Parslet rules for failing scenarios.

1. mn-samples-iec
2. mn-samples-itu
3. mn-samples-bipm
4. mn-samples-ieee-private
5. mn-samples-iho
6. mn-samples-iso
7. mn-samples-ogc
8. mn-samples-jcgm
9. mn-samples-nist


lib/plurimath/math/function/mod.rb
```diff
@@ -7,8 +7,8 @@ module Math
     module Function
       class Mod < BinaryFunction
         def to_asciimath
-          first_value = parameter_one.to_asciimath if parameter_one
-          second_value = parameter_two.to_asciimath if parameter_two
+          first_value = wrapped(parameter_one) if parameter_one
+          second_value = wrapped(parameter_two) if parameter_two
```
HassanAkbar, line 11: The first line in the `wrapped` function is `return "" unless field`. Do we need an `if` condition here?

lib/plurimath/math/function/unary_function.rb
```diff
@@ -16,30 +16,57 @@ def ==(object)
...
         def to_latex
-          first_value = "{#{parameter_one.to_latex}}" if parameter_one
+          first_value = "{#{latex_value}}" if parameter_one
           "\\#{class_name}#{first_value}"
         end
 
+        def latex_value
+          case parameter_one
+          when Array
+            parameter_one&.map { |value| value&.to_latex }.join
```
HassanAkbar, line 58:
```suggestion
            parameter_one&.compact&.map(&:to_latex)&.join
```

lib/plurimath/utility.rb
```diff
@@ -110,7 +108,7 @@ def rpr_element(wi_tag: false)
       end
 
       def update_nodes(element, nodes)
-        nodes.each { |node| element << node unless node.nil? }
+        nodes&.each { |node| element << node unless node.nil? }
```
HassanAkbar, line 111:
```suggestion
        nodes&.compact&.each { |node| element << node }
```

HassanAkbar (changes requested):

lib/plurimath/utility.rb
suleman-uzair: *`nodes`* variable contains **Ox::Element** object which supports *`each`* but doesn't support *`compact`*.

spec/plurimath/latex/metanorma_exampes_spec.rb
```diff
@@ -0,0 +1,199 @@
+require_relative "../../../lib/plurimath/math"
+
```
ronaldtse, line 2: The filename of this file contains a typo in `exampes`

spec/plurimath/latex/metanorma_exampes_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+RSpec.describe Plurimath::Latex::Parser do
```
ronaldtse, line 3: @suleman-uzair for all imported specs/tests, can you label the source of the formula? e.g. `metanorma/mn-samples-iso:{path name}`?

ronaldtse (changes requested):

spec/plurimath/latex/metanorma_exampes_spec.rb
ronaldtse: Thanks, I've also updated it.

ronaldtse:
Thank you @suleman-uzair ! Could you help resolve the remaining comments from @HassanAkbar and then merge this?

ronaldtse (approved):

PR #102, opened 2023-01-09
LatexMath specs and LATEXML conversion reference

### LINKED ISSUE #87: Copy rspec from latexmath gem, including the setup for checking against LaTeXML
1. In the latexmath gem we created a number of specs that need to be moved over here.
2. In the latexmath gem, we use the LaTeXML Perl module as the "reference converter" for LaTeX math to MathML. i.e. the output of latexmath conversion should always be identical to LaTeXML. We need to move those tests over also from the latexmath gem.

suleman-uzair:
Added LatexMath specs from [latexmath gem](https://github.com/plurimath/latexmath).
Spec's added from [_aggregator_spec.rb_](https://github.com/plurimath/latexmath/blob/main/spec/aggregator_spec.rb) file


lib/plurimath/latex/parse.rb
```diff
@@ -128,68 +100,125 @@ class Parse < Parslet::Parser
...
         end
       end
+
+      def hash_to_expression(arr)
+        type = arr.first.class
+        @@expression ||= arr.reduce do |expression, expr_string|
+          expression = dynamic_rules(expression) if expression.is_a?(type)
+          expression | dynamic_rules(expr_string)
+        end
+      end
```
HassanAkbar, line 184: The parameter is a hash so the name should reflect that.

ronaldtse (approved):

ronaldtse:
@suleman-uzair please feel free to merge when ready. Thanks!

ronaldtse:
We will still want to incorporate latexml for incorporating future latexml processing changes of the same formula, so we need to add the processing for specs.

PR #111, opened 2023-02-22
Comparison specs for MathML from Plurimath and LatexMath outputs

### LINKED ISSUE #87: Copy rspec from latexmath gem, including the setup for checking against LaTeXML

suleman-uzair:
This Pull Request _adds and updates_ specs, conversion methods, and all the necessary changes required for **Latex** to **MathML** conversion.

The **Latex** to **MathML** conversions are validated by comparing the output of [**Latexmath** gem](https://github.com/plurimath/latexmath)'s **MathML** output.

closes #87

### DIFF
--- a/lib/plurimath/asciimath/parse.rb
+++ b/lib/plurimath/asciimath/parse.rb
@@ -53,7 +53,7 @@ module Plurimath
 
       rule(:left_right) do
         (str("left") >> left_right_open_paren.as(:left) >> iteration.as(:left_right_value) >> str("right") >> left_right_close_paren.as(:right)) |
-          (table.as(:numerator) >> space.maybe >> match(/(?<!\/)\/(?!\/)/) >> space.maybe >> iteration.as(:denominator)).as(:frac) |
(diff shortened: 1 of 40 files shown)


ronaldtse:
> The **Latex** to **MathML** conversions are validated by comparing the output of [**Latexmath** gem](https://github.com/plurimath/latexmath)'s **MathML** output.

We want to move over the latexmath's LaTeXML test pipeline because we will deprecate the latexmath gem once plurimath is done. Can you do that?

suleman-uzair:
> > The **Latex** to **MathML** conversions are validated by comparing the output of [**Latexmath** gem](https://github.com/plurimath/latexmath)'s **MathML** output.
> 
> We want to move over the latexmath's LaTeXML test pipeline because we will deprecate the latexmath gem once plurimath is done. Can you do that?

Sure, I’ll do that after finalizing this PR.

ronaldtse:
Thank you @suleman-uzair !

PR #112, opened 2023-03-08
[WIP] Parslet parser rules for UnitsML

suleman-uzair:
[WIP] Added **Parslet** parser rules for **UnitsML**.

### DIFF
--- a/lib/plurimath/unitsml/constants/dimensions.yaml
+++ b/lib/plurimath/unitsml/constants/dimensions.yaml
@@ -1,0 +1,859 @@
+---
+NISTd1:
+  length:
+    powerNumerator: 1
+    symbol: L
+    dim_symbols:
+    - id: "dim_L"
+      ascii: "L"
+      html: "&#x1D5AB;"
+      mathml: "<mi mathvariant='sans-serif'>L</mi>"
+      latex: \ensuremath{\mathsf{L}}
+      unicode: "𝖫"
+
+
+NISTd2:
+  mass:
+    powerNumerator: 1
+    symbol: M
+    dim_symbols:
+    - id: "dim_M"
+      ascii: "M"
+      html: "&#x1D5AC;"
+      mathml: "<mi mathvariant='sans-serif'>M</mi>"
+      latex: \ensuremath{\mathsf{M}}
+      unicode: "𝖬"
+
+NISTd3:
+  time:
+    powerNumerator: 1
+    symbol: T
+    dim_symbols:
+    - id: "dim_T"
+      ascii: "T"
+      html: "&#x1D5B3;"
+      mathml: "<mi mathvariant='sans-serif'>T</mi>"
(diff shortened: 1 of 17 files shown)


lib/plurimath/unitsml/constants/dimensions.yaml
```diff
@@ -0,0 +1,859 @@
+---
```
ronaldtse, line 1: @suleman-uzair please note that these files are meant to be updated in the UnitsML project, so it would be best if there is a rake task to update them.

lib/plurimath/unitsml/parse.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      def load_yaml(filepath, glob_var)
+        glob_var = "@@#{glob_var}"
+        if variable_defined?(glob_var)
+          get_saved_value(glob_var)
+        else
+          content = YAML.load_file(File.join(__dir__, filepath)).to_json
+          self.class.class_variable_set(
+            glob_var,
+            JSON.parse(content, symbolize_names: true)
```
ronaldtse, line 24: Why do we convert YAML into JSON?

ronaldtse (changes requested):

lib/plurimath/unitsml/constants/dimensions.yaml
suleman-uzair: This is just for testing, I’m thinking of adding the **unitsML-db** repo as a submodule. Should I do that or add a rake task? What do you suggest?

lib/plurimath/unitsml/parse.rb
suleman-uzair: This is just for converting the keys from string to symbols, I’ll be moving this to a separate method something like `symbolize_keys`, and will be using a different approach there.

lib/plurimath/unitsml/constants/dimensions.yaml
ronaldtse: A submodule with a tagged version is good.

lib/plurimath/unitsml/parse.rb
suleman-uzair: we already have **[unitsml-ruby](https://github.com/unitsml/unitsml-ruby)  gem**  in progress, should we use that for this purpose?

ronaldtse: Yes please! The issue is https://github.com/unitsml/unitsml-ruby/issues/7

suleman-uzair:
@ronaldtse 

As per our discussion, I have completed the Unitsml conversion in the unitsml-ruby gem. However, I am not sure about the purpose of this particular pull request. Would you recommend closing it?

ronaldtse:
@suleman-uzair yes let's close this. Instead we can enhance the plurimath gem to support unitsml syntax by delegating to the unitsml-ruby gem.

PR #117, opened 2023-03-20
Fixes/asciimath missed prime

### LINKED ISSUE #116: Missed prime
`X'` in Asciimath is a convenience notation for X′, since of course ′ ″ are not Unicode characters.

This behaviour is not explicitly documented in asciimath.org, but it is in the examples:

`f'(x) = dy/dx` 

maps to:

````
<math xmlns="http://www.w3.org/1998/Math/MathML">
   <mstyle displaystyle="true">
     <mi>f</mi>
     <mo>&#x2032;</mo>
     <mrow>
       <mo>(</mo>
       <mi>x</mi> 
      <mo>)</mo>
     </mrow>
     <mo>=</mo>
     <mfrac>
       <mrow>
         <mi>d</mi>
         <mi>y</mi>
       </mrow>
       <mrow>
         <mi>d</mi>
         <mi>x</mi>
       </mrow>
     </mfrac>
   </mstyle>
 </math>
````

This mapping is in asciidoc/asciimath

https://github.com/asciidoctor/asciimath/blob/e52761299166019df0c6a1f5e714395c1e859530/lib/asciimath/parser.rb#L327

Puzzlingly, double prime is nowhere to be found. I think we will need to do it eventually anyway, and that `''` (two apostrophes) are safer for double prime than `"` (straight quote)

This is in a metanorma-standoc test case.

### PR DESCRIPTION
**Pull request by suleman-uzair**
Updated support for **'** _(prime)_ in **asciimath**.
Added support for  **''** _(double prime)_ by adding two **prime**'s without spaces between them(For **AsciiMath** only).

closes #116

### CODE REVIEW

File: lib/plurimath/asciimath/constants.rb [hunk 1]
```diff
@@ -166,6 +166,7 @@ class Constants
         "<=": :"&#x2264;",
         "|~": :"&#x2308;",
         "/_": :"&#x2220;",
+        "''": :"&#x22;",
```
opoudjis, line 169: No, sorry, Double Prime is `&#x2033;`

opoudjis (changes requested):

spec/plurimath/latex/parser_spec.rb
```diff
@@ -9501,7 +9501,17 @@
           Plurimath::Math::Symbol.new("*"),
           Plurimath::Math::Symbol.new("x"),
           Plurimath::Math::Unicode.new("&#x3e;"),
-          Plurimath::Math::Function::Text.new("\"tolerance\""),
+          Plurimath::Math::Symbol.new("\""),
```
ronaldtse, line 9504: @suleman-uzair what does this change mean?

suleman-uzair: While making changes to constants for `double prime` parsing, I realized that latex doesn't support double prime for text wrapping like AsciiMath so I updated related parser rules and specs.

opoudjis:
I have nothing to add.

spec/plurimath/latex/parser_spec.rb
suleman-uzair: According to my testing in Latexmath gem and in latex editor `\text{…} `is not the same as `"..."` so that is why I changed this.
But if you want I can revert these changes. What do you suggest?

suleman-uzair: Screenshot from 2023-03-22 13-12-15
Screenshot from 2023-03-22 13-16-16
Screenshot from 2023-03-22 13-16-18

Here in both screenshots of **AsciiMath**, we can see the `double quotes` convert math equations to simple text.
But in `Latex`'s editor, `double quotes` make no difference but the `\text` function keeps the spaces of the text and makes it readable.

ronaldtse: I am fine with this parsing, but the point is about the formula itself not being accurate. The "\text" portion should be specified in the formula. Can we add a formula to test the "\text{tolerance}" string, maybe in a new PR? Thanks.

ronaldtse (approved):

PR #120, opened 2023-03-30
Added semantics class and updated transform rules

suleman-uzair:
Added **Semantics** and **Annotation** tags support for **MathML**.
Updated **Parsed tree** structure to ignore not supported tag's attributes.


lib/plurimath/math/function/semantics.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        def to_omml_without_math_tag
+          parameter_one&.to_omml_without_math_tag
+        end
+
+        protected
+
+        def other_tags(array)
+          contented = []
+          array.map do |hash|
```
ronaldtse, line 38: This usage of a carrying array within a map method is incorrect. Map already returns the array. If you want to iterate through an array, use "each" not "map".

lib/plurimath/mathml/parser.rb
```diff
@@ -37,6 +37,12 @@ def parse_nodes(nodes)
           end
         end
       end
+
+      def validate_attributes(attributes)
+        attrs = %w[columnlines mathvariant mathcolor notation close open].freeze
```
ronaldtse, line 42: If these are always constant, define them outside of the method as class constants.

ronaldtse (changes requested):

lib/plurimath/math/function/semantics.rb
```diff
@@ -0,0 +1,52 @@
...
+          parameter_one&.to_html
+        end
+
+        def to_omml_without_math_tag
+          parameter_one&.to_omml_without_math_tag
+        end
+
+        protected
+
+        def other_tags(array)
```
ronaldtse, line 36: This method is not tested by any spec right now. When I removed the contents of this method, all specs still pass.

ronaldtse (changes requested):

lib/plurimath/math/function/semantics.rb
ronaldtse: Confirm fixed. Thanks!

ronaldtse (approved):

ronaldtse:
@suleman-uzair thank you! Can you please help bump and release?

PR #166, opened 2023-07-04
Updated omml fixtures files, conversion, and parser methods

### LINKED ISSUE #164: Matrix unexpected rendering in DOC output
*In relation to https://github.com/metanorma/pdfa-iso-32000-2/issues/5 and https://github.com/metanorma/metanorma-standoc/issues/791*

```adoc
[stem%unnumbered]
++++
[(a,b,0),(c,d,0),(e,f,1)]
++++
````

Generated DOC:
issue13-generated-doc

Original:
issue13-original

### PR DESCRIPTION
**Pull request by suleman-uzair**
This PR updates Table parsing, transformation, and conversion methods in **OMML**.

closes #164

### DIFF
--- a/lib/plurimath/math/function/table.rb
+++ b/lib/plurimath/math/function/table.rb
@@ -75,7 +75,7 @@ module Plurimath
           if value.map { |d| d.parameter_one.length == 1 }.all?
             single_td_table
           else
-            multiple_td_table
+            fenced_table(multiple_td_table)
           end
         end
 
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

File: lib/plurimath/math/function/table.rb [hunk 1]
```diff
@@ -210,6 +210,30 @@ def norm_table(table_tag)
...
+          e_node = Utility.ox_element("e", namespace: "m")
+          e_node << ox_table
+          Utility.update_nodes(d_node, [mdpr_node, e_node])
+        end
+
+        def mdpr_node
+          begchr_attr = { "m:val": open_paren } if open_paren
+          endchr_attr = { "m:val": close_paren } if close_paren
+          sepchr_attr = { "m:val": "" }
+          begchr = Utility.ox_element("begChr", attributes: begchr_attr, namespace: "m")
```
ronaldtse, line 227: I find it strange that these variables may not even be set when being used here. There must be better ways to do this...

lib/plurimath/omml/parser.rb
```diff
@@ -32,11 +32,20 @@ def parse_nodes(nodes)
                 value: parse_nodes(node.nodes),
               },
             }
+          elsif ["mr", "eqArr"].include?(node.name)
```
ronaldtse, line 35: Is this an opportunity to use case / when?

ronaldtse (changes requested):

ronaldtse (approved):

PR #169, opened 2023-07-11
Added support to allow user to input 'displaystyle' attribute value

### LINKED ISSUE #168: Do not set displaystyle=true universally
Unearthed in https://github.com/metanorma/metanorma-standoc/issues/763

Plurimath conversion to MathML is setting displaystyle=true universally:

https://github.com/plurimath/plurimath/blob/28df9c86359ae5943107997f81b65bbf07b61ae1/lib/plurimath/math/formula.rb#L23

It should not: the consumer should have the option of setting this, because they know whether the math expression occurs as a standalone formula, or in a paragraph. (In the latter case, displaystyle=false is needed.)

I suggest that .to_mathml take a displaystyle value as a parameter, defaulting to true.

suleman-uzair:
This PR implements the changes to allow users to explicitly set `displaystyle` attribute's value in **MathML**.

closes #168

### DIFF
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -3,12 +3,17 @@
 module Plurimath
   module Math
     class Formula < Core
-      attr_accessor :value, :left_right_wrapper
+      attr_accessor :value, :left_right_wrapper, :displaystyle
 
-      def initialize(value = [], left_right_wrapper = true)
+      def initialize(
+        value = [],
+        left_right_wrapper = true,
+        displaystyle: true
+      )
         @value = value.is_a?(Array) ? value : [value]
         left_right_wrapper = false if @value.first.is_a?(Function::Left)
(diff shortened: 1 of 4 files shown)


lib/plurimath/math/formula.rb
```diff
@@ -20,12 +25,12 @@ def to_asciimath
         value.map(&:to_asciimath).join(" ")
       end
 
-      def to_mathml
+      def to_mathml(display_style: displaystyle)
```
HassanAkbar, line 28: @suleman-uzair can you add a test case for this?

ronaldtse:
Ping @suleman-uzair is this ready to merge? Thanks.

suleman-uzair:
> Ping @suleman-uzair is this ready to merge? Thanks.

Yes, It's ready.

opoudjis (approved):

PR #196, opened 2023-11-08
Opal compatibility + Oga support

hmdne:
Opal is a Ruby-to-JavaScript transpiler that involves certain semantics that the compiled code has to follow. Most importantly, those are as follows:

* No C library extensions
* Assumption that strings are immutable
* Correct knowledge about required files before compilation takes place

Fixing those issues in code will allow us to compile Plurimath to JavaScript, so that the code will be usable directly in web browsers and environments like NodeJS. This is the work that we are doing in the `plurimath-js` repository and therefore any issues related to the JavaScript version of the library should be directed there.

One of the most important parts is that the Ox library, as of now, requires a C extension. An approach we took was to port another library that does mostly the same, namely Oga and then to port Plurimath to support both Ox and Oga. We also chose Oga, because of apparent support of Oga, along with Nokogiri, in the `equivalent-xml` library, which we use for comparing XML trees.

Support for both has been implemented by doing a façade pattern named `Plurimath.xml_engine` that uses Ox directly if needed and Oga via an API that is mostly compatible with Ox (as per usage patterns in Plurimath codebase).

The Ruby runtime still uses Ox, except when an environment variable `PLURIMATH_OGA` is set, in which case, we use Oga. We use this implementation internally to test for Oga shim completeness while still running code on MRI. If problems arise, this means that it's not Opal fault directly, that problems happen, but Oga shim issues.

Other changes most likely refer to Opal build system, minor compatibility fixes (like assumptions that strings are mutable), corrections of paths. All of those changes should keep the logic mostly the same when running in MRI.

This is not a full solution for Opal compatibility yet. There are still important parts of that work in progress. It's not expected at this point that either all Opal or Oga tests will pass. It is expected though that all the tests that are currently being ran will pass.

Ref: plurimath-js/plurimath-js#3

### DIFF
--- a/lib/plurimath/math.rb
+++ b/lib/plurimath/math.rb
@@ -19,9 +19,7 @@ require_relative "latex/parser"
 require_relative "html/parser"
 require_relative "omml/parser"
 require_relative "utility"
-require "ox"
 require "yaml"
-Ox.default_options = { encoding: "UTF-8" }
 
 module Plurimath
   module Math
--- a/lib/plurimath/xml_engine/oga.rb
+++ b/lib/plurimath/xml_engine/oga.rb
@@ -1,0 +1,246 @@
+# frozen_string_literal: true
+
+require "plurimath/xml_engine"
+require "corelib/array/pack" if RUBY_ENGINE == "opal"
+require "oga"
+
+module Plurimath
+  module XMLEngine
+    class Oga
+      class << self
+        def new_element(name)
+          data = ::Oga::XML::Element.new(name: name)
+          Node.new(data)
+        end
+
+        def dump(data, indent: nil)
+          Dumper.new(data, indent: indent).dump.out
+        end
+
+        def load(data)
+          data = ::Oga::XML::Parser.new(data, html: true).parse
+          if data.xml_declaration
+            Document.new(data)
+          else
+            Document.new(data).nodes.first
+          end
+        end
+
+        def is_xml_comment?(node)
+          node = node.unwrap if node.respond_to? :unwrap
+          node.is_a?(Comment)
+        end
+      end
+
+      # Create API compatible with Ox, per Plurimath usage
+      class Wrapper
+        def initialize(value)
+          @wrapped = value
+        end
+
+        def unwrap
+          @wrapped
+        end
+
+        def ==(other)
+          self.class == other.class &&
+            @wrapped.inspect == other.unwrap.inspect
+        end
+      end
+
+      class Node < Wrapper
+        # Ox removes text nodes that are whitespace-only.
+        # There exists a weird edge case on which Plurimath depends:
+        # <mi> <!-- xxx --> &#x3C0;<!--GREEK SMALL LETTER PI--> </mi>
+        # If the last text node of an element that does not contain other
+        # elements is a whitespace, it preserves it. The first one can be
+        # safely removed.
+        def nodes
+          children = @wrapped.children
+          length = children.length
+          preserve_last = true
+          children.map.with_index do |i,idx|
+            if preserve_last && idx == length-1 && i.is_a?(::Oga::XML::Text)
+              i.text
+            elsif i.is_a? ::Oga::XML::Text
+              remove_indentation(i)
+            elsif i.is_a? ::Oga::XML::Comment
+              Node.new(i)
+            else
+              preserve_last = false
+              Node.new(i)
+            end
+          end.compact
+        end
+
+        def [](attr)
+          attr = attr.to_s
+
+          @wrapped.attributes.each do |e|
+            return e.value if [e.name, e.name.split(":").last].include? attr
+          end
+
+          nil
+        end
+
+        def []=(attr, value)
+          # Here we tap into the internal representation due to some likely
+          # bug in Oga
+          attr = ::Oga::XML::Attribute.new(name: attr.to_s)
+          attr.element = @wrapped
+          attr.instance_variable_set(:@value, value.to_s)
+          attr.instance_variable_set(:@decoded, true)
+          @wrapped.attributes << attr
+        end
+
+        def <<(other)
+          other = other.unwrap if other.respond_to? :unwrap
+
+          case other
+          when String
+            text = other
+            # Here we tap into the internal representation due to some likely
+            # bug in Oga
+            other = ::Oga::XML::Text.new
+            other.instance_variable_set(:@from_plurimath, true)
+            other.instance_variable_set(:@text, text)
+            other.instance_variable_set(:@decoded, true)
+          end
+
+          @wrapped.children << other.dup
+          self
+        end
+
+        def attributes
+          @wrapped.attributes.to_h do |e|
+            [e.name.split(":").last, e.value]
+          end
+        end
+
+        def locate(xpath)
+          @wrapped.xpath(xpath).map do |i|
+            case i
+            when ::Oga::XML::Text
+              i.text
+            when ::Oga::XML::Attribute
+              i.value
+            else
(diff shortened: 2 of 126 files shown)


ronaldtse (approved):

ronaldtse:
Thank you @hmdne , let's have @suleman-uzair and @HassanAkbar review this before merging.

Please also help fix the tests before this is mergable, thanks.

hmdne:
> Please also help fix the tests before this is mergable, thanks.

The tests are passing, it looks like a CI failure.

ronaldtse:
@suleman-uzair could you please help fix the "tests-passed" test since it doesn't seem to be related to the changes here? Thanks.

suleman-uzair:
@ronaldtse I will look into it as soon as possible.
@hmdne, could you please add tests for the new code added to resolve the code coverage error?

lib/plurimath/math.rb
```diff
@@ -19,9 +19,8 @@
 require_relative "html/parser"
 require_relative "omml/parser"
 require_relative "utility"
-require "ox"
+
```
suleman-uzair, line 22: This seems to be an extra and unnecessary space.

lib/plurimath/xml_engine/oga.rb
```diff
@@ -0,0 +1,231 @@
...
+
+        attr_reader :out
+
+        private
+
+        def dump_attrs(node)
+          node.attributes.map do |i|
+            #if i.namespace
+            #  %{ #{i.namespace.name}:#{i.name}="#{attr_entities i.value}"}
+            #else
```
suleman-uzair, line 178: Unused code should be removed.

spec/spec_helper.rb
```diff
@@ -1,11 +1,15 @@
...
+  SimpleCov.start do
+    add_filter "/spec/"
+  end
 
-require "bundler/setup"
+  require "bundler/setup"
+end
 require "plurimath"
 require "rspec/matchers"
+# require 'nokogiri'
```
suleman-uzair, line 11: let's remove this line if we don't need it.

lib/plurimath/xml_engine/oga.rb
```diff
@@ -0,0 +1,231 @@
...
+              "&lt;"
+            elsif i == ORD_GT
+              "&gt;"
+            elsif i == ORD_APOS && attr
+              "&apos;"
+            elsif i == ORD_QUOT && attr
+              "&quot;"
+            elsif i == ORD_NEWLINE || i == ORD_CARRIAGERETURN
+              i.chr("utf-8")
+            elsif i == ORD_AMP || i == ORD_LT || i == ORD_GT ||
```
suleman-uzair, line 206: shouldn't case be used instead of if?

hmdne:
Thank you for the review, @suleman-uzair. I will come back to this tomorrow.

As of now, I have one question and perhaps you will be one who knows the answer to that:

I am dealing with this testcase: `rspec ./spec/plurimath/omml_spec.rb:60 # Plurimath::Omml.to_latex, to_mathml, to_asciimath contains example #01 returns parsed Asciimath to Formula` - in particular with MathML output.

The diff I receive, where `-` is correct, `+` is what my (Opal) port outputs.

```diff
        <math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
          <mstyle displaystyle="true">
            <msubsup>
       -      <mo>&#x222b;</mo>
       +      <mo>int</mo>
              <mn>2</mn>
              <mn>1</mn>
            </msubsup>
```

I have tried to investigate this problem thoroughly. What I found is this. The difference stems from this particular function:

https://github.com/plurimath/plurimath/blob/main/lib/plurimath/math/symbol.rb#L28-L40

In particular, this fragment:

```ruby
          mo_value = (unicodes[value] || unicode || value).to_s
```

In the case of Opal, `unicodes[value]` is truthy, so the final result is `int`. In case of Ruby, this is falsy, so the final result is `&#x222b;`.

 value is `"&#x222b;"`. Let's find this Hash.

https://github.com/plurimath/plurimath/blob/main/lib/plurimath/mathml/constants.rb#L96

Ok - so it's quite simple why this is falsy in Ruby. `"&#x222b;": "int",` is equivalent to `        :"&#x222b;" => "int",`. Symbols are not Strings, there's only Symbol available, there's no String with such a name in this hash, so we get to return `value` which is `"&#x222b;"` and test passes.

In Opal, Strings are the same type as Symbols. So, incorrectly, this is truthy and the value of the entire expression is `"int"`, so the test fails.

I have re-ran the test suite using Ruby (not Opal) and it seems like, `value` is always a `String`, never a `Symbol` (for all test cases at least). Therefore if I just patch the file `lib/plurimath/math/symbol.rb` this way...:

```diff
diff --git a/lib/plurimath/math/symbol.rb b/lib/plurimath/math/symbol.rb
index 70f8b37..7f81670 100644
--- a/lib/plurimath/math/symbol.rb
+++ b/lib/plurimath/math/symbol.rb
@@ -32,7 +32,7 @@ module Plurimath
         unicodes = Mathml::Constants::UNICODE_SYMBOLS
         unicode = unicodes.invert[value]
         if operator?(unicode) || unicode || explicit_checks(unicode)
-          mo_value = (unicodes[value] || unicode || value).to_s
+          mo_value = (unicode || value).to_s
           return Utility.ox_element("mo") << mo_value
         end
 ```

...the Ruby-side stays equivalent and at the same time Opal passes the test.

I have a suspicion though, that this isn't what was intended. In fact, I suppose that the logic in this function may be incorrect, and what was supposed to happen was this..., at least (if not even the `unicode` variable should work in a different way):

```diff
diff --git a/lib/plurimath/math/symbol.rb b/lib/plurimath/math/symbol.rb
index 70f8b37..7f81670 100644
--- a/lib/plurimath/math/symbol.rb
+++ b/lib/plurimath/math/symbol.rb
@@ -32,7 +32,7 @@ module Plurimath
         unicodes = Mathml::Constants::UNICODE_SYMBOLS
         unicode = unicodes.invert[value]
         if operator?(unicode) || unicode || explicit_checks(unicode)
-          mo_value = (unicodes[value] || unicode || value).to_s
+          mo_value = (unicodes[value.to_sym] || unicode || value).to_s
           return Utility.ox_element("mo") << mo_value
         end
 ```

...but then the tests are also incorrect. Based on the git history, you also seem to be an author of this fragment. Can you perhaps take a second look at this logic?

lib/plurimath/xml_engine/oga.rb
hmdne: There's a check for `&& attr` on line 200 and 202, so unfortunately we can't use `case`.

hmdne:
I have moved the last comment into a separate issue. Let's try to get this part merged. I will follow-up in a next comment.

hmdne:
@ronaldtse @suleman-uzair 

I have made ahead-of-time Plurimath fully work with either Ox or Oga. To run tests under Ox, you can run `rake` just like before. To run tests under Oga, you can run `PLURIMATH_OGA=1 rake`.

Additionally, as requested, I have written a test case (xml_engine_spec.rb) that regardless of a setting mentioned above, tests the XML engine boilerplate for both engines.

There are only a couple of lines that are not covered now. One is an XPath edge case that I couldn't trigger consistently for both engines and isn't used by Plurimath currently, but from a perspective of having a full implementation makes sense for me. The other ones are guarded by a `RUBY_ENGINE == 'opal'` macro, therefore they aren't ran under MRI.

HassanAkbar (approved):
Looks good to me 👍

ronaldtse (approved):
Thank you @hmdne !

PR #209, opened 2023-12-22
Added exception for non-numeric exponent value

### LINKED ISSUE #208: Exception for non-numeric exponent value for UnitsML
@suleman-uzair This becomes something you should handle, now that Plurimath encompasses unitsml

At issue: if the exponents of unit expressions in UnitsML are not numbers but variables, e.g. `unitsml(kg*m^(p)*s^(q))` instead of `unitsml(kg*m^(-2)*s^(-2))`, an error needs to be raised, to indicate that this is not a real unit, and is thus not legal to pass into unitsml.

_Originally posted by @opoudjis in https://github.com/plurimath/asciimath2unitsml/issues/32#issuecomment-1862816327_

suleman-uzair:
This PR adds error raising check for UnitsML equations containing non-numeric values in exponent.

closes #208

### DIFF
--- a/lib/plurimath/unitsml.rb
+++ b/lib/plurimath/unitsml.rb
@@ -7,10 +7,20 @@ module Plurimath
 
     def initialize(text)
       @text = text
+      raise Math::ParseError.new(error_message) if text.match?(/\^(\([^\d-])|[^\d-]\)/)
     end
 
     def to_formula
       ::Unitsml.parse(text).to_plurimath
+    end
+
+    def error_message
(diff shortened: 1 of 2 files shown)


opoudjis (approved):
Cool. I momentarily wondered whether this should be testing for U+2212 "minus", but U+2212 doesn't really belong in Asciimath at all...

ronaldtse (approved):

ronaldtse:
Thank you @suleman-uzair !

PR #221, opened 2024-02-27
Spaces before Unitsml in MathML

### LINKED ISSUE #222: Add spaces before UnitsML in MathML conversion(using Ox)
Add spaces before **Unitsml** in **MathML** output following the instructions mentioned in the https://github.com/metanorma/bipm-si-brochure/issues/222

Process **Ox** nodes in post conversion methods.

suleman-uzair:
This PR adds spaces between normal **MathML** tags and  **Unitsml** generated **MathML** as instructed [here](https://github.com/metanorma/bipm-si-brochure/issues/222#issuecomment-1966110260).

closes #222

### DIFF
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -3,7 +3,7 @@
 module Plurimath
   module Math
     class Formula < Core
-      attr_accessor :value, :left_right_wrapper, :displaystyle, :input_string
+      attr_accessor :value, :left_right_wrapper, :displaystyle, :input_string, :unitsml
 
       MATH_ZONE_TYPES = %i[
         omml
@@ -16,12 +16,14 @@ module Plurimath
         value = [],
         left_right_wrapper = true,
         display_style: true,
-        input_string: nil
+        input_string: nil,
+        unitsml: false
       )
         @value = value.is_a?(Array) ? value : [value]
         left_right_wrapper = false if @value.first.is_a?(Function::Left)
         @left_right_wrapper = left_right_wrapper
         @displaystyle = boolean_display_style(display_style)
+        @unitsml = unitsml if unitsml
       end
 
       def ==(object)
@@ -49,6 +51,7 @@ module Plurimath
         style = ox_element("mstyle", attributes: style_attrs)
         Utility.update_nodes(style, mathml_content)
         Utility.update_nodes(math, [style])
+        unitsml_post_processing(math)
         dump_nodes(math, indent: 2)
       rescue
         parse_error!(:mathml)
@@ -63,10 +66,9 @@ module Plurimath
       def to_mathml_without_math_tag
         return mathml_content unless left_right_wrapper
 
-        Utility.update_nodes(
-          Utility.ox_element("mrow"),
-          mathml_content,
-        )
+        mrow = ox_element("mrow")
+        mrow.attributes[:unitsml] = true if unitsml
+        Utility.update_nodes(mrow, mathml_content)
       end
 
       def mathml_content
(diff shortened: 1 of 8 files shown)


lib/plurimath/math/formula.rb
```diff
@@ -273,6 +275,31 @@ def omml_br_tag
         r_tag = ox_element("r", namespace: "m")
         r_tag << ox_element("br")
       end
+
+      def unitsml_post_processing(nodes)
+        nodes.each.with_index do |node, index|
+          if node.is_a?(Ox::Element) && node.attributes&.dig(:unitsml)
+            previous = nodes[index-1]
+            if previous && ["mi", "mn"].include?(previous.name)
+              if dump_ox_nodes(node.nodes).match?(/>(&#x[0-9a-fA-F]+;)*(\w+|\d+)</)
```
opoudjis, line 284: This condition isn't right. 

````
text = HTMLEntities.new.encode(Nokogiri::XML("<mrow>#{rendering}</mrow>")
        .text.strip)
      if /\p{L}|\p{N}/.match?(text)
        "<mo rspace='thickmathspace'>&#x2062;</mo>"
      else "<mo>&#x2062;</mo>"
     end
````

This is saying that if, according to unicode, there is a number or a letter in the text (`/\p{L}|\p{N}/`), use a thickmathspace; otherwise, do not. In the case of say 9°, ° is not a letter or a number, therefore there is no space between 9 and °.

But your test is using Ascii-based \w and \d, and matches on any XML escape. That means it will not pick up that Å is a letter (it's not a standard metric unit, but it is still a unit); and if ° is encoded as an escape, `&#x00b0;`, it will not pick up that ° is not a letter or number.

Please do a more sophisticated test. You need to extract the text content of `dump_ox_nodes(node.nodes)` (the equivalent of Nokogiri `nodes.text`), and apply specifically the test `/\p{L}|\p{N}/.match?(HTMLEntities.new.decode(text))` to it, resolving the XML escapes to their Unicode counterparts.

(My source code had `.encode`; but sadly, that is a bug.)

In testing, we expect to see:

* `<mn>9</mn> <mi>Å</mi>` => `<mn>9</mn> <mo space="thickspace">&#x2062;</mo><mi>Å</mi>` : _9 Å_
* `<mn>9</mn> <mi>°</mi>` => `<mn>9</mn> <mo>&#x2062;</mo><mi>°</mi>`: _9°_


You  currently have a test on `'1 "unitsml(um)"'`. Add tests on `'1 "unitsml(Aring)"'` and `'1 "unitsml(deg)"'`. (Symbols documented on https://github.com/plurimath/asciimath2unitsml )

Btw, the symbols documentation appears in Plurimath.org, right?

opoudjis (changes requested):

suleman-uzair:
@opoudjis
I have updated the regex condition as you mentioned.
> Btw, the symbols documentation appears in [Plurimath.org](http://plurimath.org/), right?

**Plurimath.org** doesn’t have the list of symbols for **Unitsml** yet.
But **LaTeX**, **AsciiMath**, and **MathML** supported symbols list are available.

opoudjis (approved):

opoudjis:
> Plurimath.org doesn’t have the list of symbols for Unitsml yet.
> But LaTeX, AsciiMath, and MathML supported symbols list are available.

Should I create this as a ticket?

suleman-uzair:
> Should I create this as a ticket?

@opoudjis Yes sure, I will work on it after finishing current task.

suleman-uzair:
@opoudjis 
I have created the issue [plurimath/plurimath.github.io#31](https://github.com/plurimath/plurimath.github.io/issues/31).

lib/plurimath/math/formula.rb
ronaldtse: @opoudjis can we merge this now? Thanks.

opoudjis: Yes, I reviewed and approved

PR #243, opened 2024-04-23
Number Formatter

### LINKED ISSUE #232: Create number formatter in Plurimath
In https://github.com/metanorma/iso-iec-80000/issues/39 the desire was expressed to move number formatting code from Metanorma to Plurimath. The code is 

https://github.com/metanorma/isodoc/blob/main/lib/isodoc/presentation_function/math.rb : localized_number(), and probably twitter_cldr_localiser() (which draws on [twitter-cldr-rb](https://github.com/twitter/twitter-cldr-rb) for its definition of number localisation, and expands on it).

Note that flavours customise twitter_cldr_localiser_symbols(), to define default rendering for the Metanorma flavour, overriding the values given by twitter-cldr-rb for the current locale. For example,

https://github.com/metanorma/isodoc/blob/2d650f7788abf0eb14d89858c7877c3e36f927d3/lib/isodoc/presentation_function/math.rb#L65

````
def twitter_cldr_localiser_symbols
        { group: "'" }
end
````

That tells ITU to break up groups of three digits with "'" not space.

https://github.com/metanorma/metanorma-bipm/blob/4aab355394d6a6b18ff884100627f3d102548ffb/lib/isodoc/bipm/presentation_xml_convert.rb#L77

````
      def twitter_cldr_localiser_symbols
        { group: "&#xA0;", fraction_group: "&#xA0;",
          fraction_group_digits: 3 }
      end
````

That tells BIPM to separate digits and fraction digits, every 3 digits, with nbsp. I want to pass the hash in twitter_cldr_localiser_symbols to the class realising this number formatter, since they are defaults for the entire run.

I am requesting that you integrate this code into Plurimath, so long as I can call it cleanly: it will be a different class of code than the Plurimath code conversions, and it will take a numeric string as input, not a MathML node. Note that localize_maths() in that code is actually iterating through mn, and calculating what the precision of the number is if it isn't given. You may end up doing that in general for all rendered MathML, but don't do it yet. 

I'm only going to give you a numeric string, and ask you to calculate the precision if it is not given. I also want you to generate the parameters which are being assumed in my code, locale, twitter_cldr_reader, and twitter_cldr_localiser_symbols.

I am THEN asking you to expand this method to do scientific notation. But one thing at a time.

This is how I am currently invoking localized_number(). Note that precision is derived if not given; and that twitter_cldr_reader has already been initialised from locale before running this code.

````
def localize_maths(node, locale)
      node.xpath(".//m:mn", MATHML).each do |x|
        num = BigDecimal(x.text)
        precision = /\./.match?(x.text) ? x.text.sub(/^.*\./, "").size : 0
        x.children = localized_number(num, locale, precision)
      rescue ArgumentError
      end
    end
````

I want you to realise the following code, integrating my code:

````
require "twitter-cldr-rb"

class Plurimath::Numberformatter
def initialize(locale_str: "en", twitter_cldr_localiser_symbols: {})
  @locale,  @twitter_cldr_reader = twitter_cldr_localiser(locale_str)
  @twitter_cldr_localiser_symbols = twitter_cldr_localiser_symbols
end

    def twitter_cldr_reader(locale)
      return @twitter_cldr_reader if @twitter_cldr_reader

      num = TwitterCldr::DataReaders::NumberDataReader.new(locale)
      num.symbols.merge!(@twitter_cldr_localiser_symbols)
        .merge!(parse_localize_number)
    end

    def twitter_cldr_localiser(locale_str: "en")
      locale = TwitterCldr.supported_locale?(locale_str.to_sym) ? locale_str.to_sym : :en
      [locale , twitter_cldr_reader(locale)]
    end

  def parse_localize_number
  # taken from my code: these are the default values of the various parameters determining how the number formatter behaves
 end

# note: I have not defined digitcount yet
def localized_number(number_string, locale: "en", precision: nil, digitcount: nil, format: {})
   locale_sym, reader = if locale == @locale.to_s
     [@locale,  @twitter_cldr_reader]
   else
     twitter_cldr_reader(locale)
   end
   num = BigDecimal(number_string.to_s) # we need arbitrary precision
   precision ||= /\./.match?(x.text) ? x.text.sub(/^.*\./, "").size : 0 # if we don't know the precision, infer it
   reader.merge!(format)
   localized_number(num, locale_sym, reader, precision) # from this point use my existing implementation, but with twitter_cldr_reader now an explicit parameter, in case we switch it mid-run
end

end
````

As a result, I expect to be able to make the following calls 

````
n = Plurimath::Numberformatter.new("de", { group: "&#xA0;", fraction_group: "&#xA0;",
          fraction_group_digits: 3 }) 
n.localized_number(1) # 1
n.localized_number("1") # 1
n.localized_number(1.6, locale: "en") # 1.6
n.localized_number(1.6, locale: "fr", precision: 3) # 1,600 : the locale "fr" is resulting in twitter_cldr_reader reading in French localisation of numbers, according to which the parameter "decimal", delimiting decimal numbers, is "," not "."
n.localized_number(1.6, locale: "fr", precision: 3, format: {decimal: "*"} ) # 1*600 : override the locale parameter "decimal" with my own
````

### PR DESCRIPTION
**Pull request by suleman-uzair**
This PR adds a **NumberFormatter** class to process numeric strings.

closes #232

### CODE REVIEW

File: lib/plurimath/formatter/number_formatter.rb [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+
+module Plurimath
+  module Formatter
+    class NumberFormatter
+      attr_accessor :locale, :localize_number, :localiser_symbols
+
+      LOCALIZE_NUMBER_REGEX = %r{(?<group>[^#])?(?<groupdigits>#+0)(?<decimal>.)(?<fractdigits>#+)(?<fractgroup>[^#])?}
+      SUPPORTED_NOTATIONS = %i[e scientific engineering].freeze
+
+      def initialize(locale, localize_number:, localiser_symbols:)
```
HassanAkbar, line 14: I think we should be using `s` or `z` in localize everywhere in plurimath.
 Having `localize_number` along with `localiser_symbol` does not feels right to me.

lib/plurimath/formatter/twitter_cldr_rb/fraction.rb
```diff
@@ -0,0 +1,57 @@
...
+        protected
+
+        def change_format(precision, string)
+          tokens = []
+          tokens << string&.slice!(0, group) until string&.empty?
+          tokens.compact.join(separator)
+        end
+
+        def digit_count_format(int, fraction, number)
+          integer = (int + "." + fraction)
```
HassanAkbar, line 43:
```suggestion
          integer = int + "." + fraction
```

lib/plurimath/formatter/number_formatter.rb
```diff
@@ -0,0 +1,116 @@
...
+        m = LOCALIZE_NUMBER_REGEX.match(@localize_number) or return {}
+        ret = {
+          decimal: m[:decimal],
+          group_digits: m[:groupdigits].size,
+          fraction_group_digits: m[:fractdigits].size,
+          group: m[:group] || "",
+          fraction_group: m[:fractgroup] || ""
+        }.compact
+        %i(group fraction_group).each { |x| ret[x] == " " and ret[x] = "\u00A0" }
+        ret
```
HassanAkbar, line 49:
```suggestion
        {
          decimal: m[:decimal],
          group_digits: m[:groupdigits].size,
          fraction_group_digits: m[:fractdigits].size,
          group: m[:group] == " " ? "\u00A0" : (m[:group] || ""),
          fraction_group: m[:fractgroup] == " " ? "\u00A0" : (m[:fractgroup]  || "")
        }.compact
```

lib/plurimath/formatter/twitter_cldr_rb/fraction.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def apply(fraction, options = {}, int = "")
+          precision = options[:precision] || self.precision
+          if precision > 0
+            number = interpolate(format(options), fraction, :left)
+            number = digit_count_format(int, fraction, number) if @digit_count
+            decimal + change_format(precision, number)
+          else
+            ''
+          end
```
HassanAkbar, line 26:
```suggestion
          return "" unless precision > 0
          
          number = interpolate(format(options), fraction, :left)
          number = digit_count_format(int, fraction, number) if @digit_count
          decimal + change_format(precision, number)
```

opoudjis:
Will get to this tomorrow

opoudjis (approved):
Delighted! Thank you so much @suleman-uzair, this was a pleasure to read through!

ronaldtse:
Thanks @suleman-uzair !

PR #258, opened 2024-06-05
allow arbitrary precision in numeric formatter: closes https://github…

opoudjis:
….com/plurimath/plurimath/issues/232


suleman-uzair:
@opoudjis just released version `0.8.7`.

PR #263, opened 2024-06-25
use BigDecimal for digit_count: https://github.com/plurimath/plurimat…

opoudjis:
…h/issues/262

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,3 +12,4 @@ gem 'opal-rspec', "~> 1.1.0a"
 gem 'oga'
 gem 'ox'
 gem "unitsml"
+gem "debug"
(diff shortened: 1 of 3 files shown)


suleman-uzair:
@opoudjis will release the version as soon as checks pass.

suleman-uzair (approved):

suleman-uzair:
@opoudjis just released version **0.8.9**.

opoudjis:
Thank you!

ronaldtse:
@suleman-uzair Plurimath is also compiled to JS through Opal. Does Opal provide BigDecimal? If not we need to do a switch during compilation.

ping @hmdne

suleman-uzair:
> Does Opal provide BigDecimal? 

@ronaldtse Yes, It is mentioned in the documentation [here](https://opalrb.com/docs/api/v1.8.1/stdlib/BigDecimal.html) that **Opal** supports **BigDecimal**, So I think it should work.

ronaldtse:
That is awesome, thanks!

PR #268, opened 2024-06-28
Significant digits support

### LINKED ISSUE #264: Need significant digits
The number formatter needs to be able to specify the number of significant digits. This is particularly important for Metanorma, as I will be formatting the number by default to the number of significant digits provided, and I am ending up translating in and out of e-notation: https://github.com/metanorma/metanorma-standoc/issues/882

I have been using digit_count to that end until now. But when faced with leading zeroes, digit_count is clearly counting leading zeroes. If this is about fitting digits on a screen, I see that it needs to. But I do need the ability to specify significant digits as distinct from digit count. And (unlike what BigDecimal does) I need trailing zeroes to be considered significant.

So:

````
num1 = "0.001"
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {}) == 0,001
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:digit_count=>3}) == 0,0 # !!!
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:digit_count=>3, notation: :scientific}) == 1,0000 × 10^-03
````

(Not sure what is going on with the last one: is it counting the exponent as a significant digit, somehow? The second also looks wrong.)

What I need is:

````
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:digit_count=>3}) == 0,00 # three digits
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:digit_count=>6, notation: :scientific}) ==  1,00 × 10^-03 OR 1,00000 × 10^-03 # need to decided whether counting exponent digits or not. I think we do, and I also think we aren't at the moment....
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:significant=>3}) == 0,00100
puts Plurimath::NumberFormatter.new(:de).localized_number(num1, format: {:significant=>3, notation: :scientific}) == 1,00 × 10^-03
````

So that `100` are the three significant digits. And that  is regardless of how many digits  are actually displayed. (So I expect digit_count to include leading zeroes, and significant not to.)

This applies for fractional digits; for integers, digit_count is acting like I expect significant to.

### PR DESCRIPTION
**Pull request by suleman-uzair**
This PR
1. Updates digit_count scenario
2. Adds support of the `significant:` option for significant numbers.

closes #264

### DIFF
--- a/lib/plurimath/formatter/number_formatter.rb
+++ b/lib/plurimath/formatter/number_formatter.rb
@@ -14,6 +14,8 @@ module Plurimath
         result = integer_format.apply(int, options)
         result << fraction_format.apply(fraction, options, int) if fraction
 
+        result = sigfigs(result) unless significant.zero?
+
         number_system.transliterate(
           "#{prefix.to_s}#{result}#{suffix.to_s}"
         )
@@ -34,6 +36,62 @@ module Plurimath
             data_reader.symbols
           )
         ]
+      end
+
+      def sigfigs(str)
+        return str unless str.match?(/[1-9]/)
+
+        signify_fraction(str)
+      end
+
+      def signify_fraction(str)
+        chars = str.chars
+        sig_num = false
+        sig_count = significant
+        frac_part = false
+        new_str = chars.map.with_index do |char, ind|
+          if char == decimal_char
+            frac_part = true
+          elsif sig_count.zero?
+            next if frac_part
+
+            next "0"
+          elsif sig_num
+            sig_count -= 1
+            round_str(chars, ind, frac_part) if sig_count.zero?
+          elsif char.match?(/[1-9]/)
+            sig_num = true
+            sig_count -= 1
+          end
+          char
+        end.join
+        if sig_count > 0
+          decimal = frac_part ? "" : decimal_char
+          new_str << (decimal + ("0" * sig_count))
+        end
+        new_str
+      end
+
+      def decimal_char
+        data_reader.symbols[:decimal]
+      end
+
+      def significant
+        data_reader.symbols[:significant] || 0
+      end
+
+      def round_str(chars, index, frac_part)
+        plus_one = false
+        chars.reverse.each.with_index do |char, ind|
+          next char unless index >= ind
+
+          if char == "9" && plus_one
+          elsif plus_one
+          elsif char.match?(/[5-9]/)
+          else
+            next char
+          end
+        end
       end
     end
   end
--- a/lib/plurimath/formatter/number_formatter.rb
+++ b/lib/plurimath/formatter/number_formatter.rb
@@ -49,27 +49,27 @@ module Plurimath
         sig_num = false
         sig_count = significant
         frac_part = false
-        new_str = chars.map.with_index do |char, ind|
-          if char == decimal_char
-            frac_part = true
-          elsif sig_count.zero?
-            next if frac_part
+        new_array = []
+        chars.each.with_index do |char, ind|
+          frac_part = frac_part || char == decimal_char
+          sig_num = sig_num || char.match?(/[1-9]/)
 
-            next "0"
-          elsif sig_num
-            sig_count -= 1
-            round_str(chars, ind, frac_part) if sig_count.zero?
-          elsif char.match?(/[1-9]/)
-            sig_num = true
-            sig_count -= 1
-          end
-          char
-        end.join
-        if sig_count > 0
-          decimal = frac_part ? "" : decimal_char
-          new_str << (decimal + ("0" * sig_count))
+          break if sig_count.zero?
+
+          new_array << char
+          next unless sig_num || char.match?(/[0-9]/)
+
+          sig_count -= 1
         end
-        new_str
+        if sig_count > 0
+          new_array << decimal_char unless frac_part
+          new_array << ("0" * sig_count)
+        else
+          remain_chars = chars.length - significant
+          round_str(chars, new_array, frac_part) if remain_chars > 0
+          new_array << ("0" * remain_chars)
+        end
+        new_array
       end
 
       def decimal_char
@@ -80,7 +80,9 @@ module Plurimath
         data_reader.symbols[:significant] || 0
       end
 
-      def round_str(chars, index, frac_part)
+      def round_str(chars, new_array, frac_part)
+        return unless chars[new_array.length].match?(/[5-9]/)
+
         plus_one = false
         chars.reverse.each.with_index do |char, ind|
           next char unless index >= ind
--- a/lib/plurimath/formatter/number_formatter.rb
+++ b/lib/plurimath/formatter/number_formatter.rb
@@ -7,14 +7,14 @@ module Plurimath
         options[:precision] ||= precision_from(number)
         options[:type] ||= :decimal
 
-        prefix, suffix, integer_format, fraction_format = *partition_tokens(tokens)
+        prefix, suffix, integer_format, fraction_format, signif_format = *partition_tokens(tokens)
         number = truncate_number(number, integer_format.format.length)
 
         int, fraction = parse_number(number, options)
         result = integer_format.apply(int, options)
         result << fraction_format.apply(fraction, options, int) if fraction
 
-        result = sigfigs(result) unless significant.zero?
+        result = signif_format.apply(result, integer_format, fraction_format)
 
         number_system.transliterate(
           "#{prefix.to_s}#{result}#{suffix.to_s}"
@@ -34,66 +34,11 @@ module Plurimath
           Numbers::Fraction.new(
             tokens[1],
             data_reader.symbols
+          ),
+          Numbers::Significant.new(
+            data_reader.symbols,
           )
         ]
-      end
-
-      def sigfigs(str)
-        return str unless str.match?(/[1-9]/)
-
-        signify_fraction(str)
-      end
-
-      def signify_fraction(str)
-        chars = str.chars
-        sig_num = false
-        sig_count = significant
-        frac_part = false
-        new_array = []
-        chars.each.with_index do |char, ind|
-          frac_part = frac_part || char == decimal_char
-          sig_num = sig_num || char.match?(/[1-9]/)
-
-          break if sig_count.zero?
-
-          new_array << char
-          next unless sig_num || char.match?(/[0-9]/)
-
-          sig_count -= 1
-        end
-        if sig_count > 0
-          new_array << decimal_char unless frac_part
-          new_array << ("0" * sig_count)
-        else
-          remain_chars = chars.length - significant
-          round_str(chars, new_array, frac_part) if remain_chars > 0
-          new_array << ("0" * remain_chars)
-        end
-        new_array
-      end
-
-      def decimal_char
-        data_reader.symbols[:decimal]
-      end
-
-      def significant
-        data_reader.symbols[:significant] || 0
-      end
-
-      def round_str(chars, new_array, frac_part)
-        return unless chars[new_array.length].match?(/[5-9]/)
-
-        plus_one = false
-        chars.reverse.each.with_index do |char, ind|
-          next char unless index >= ind
-
-          if char == "9" && plus_one
-          elsif plus_one
-          elsif char.match?(/[5-9]/)
-          else
-            next char
-          end
-        end
       end
     end
   end
--- a/lib/plurimath/formatter/numbers/fraction.rb
+++ b/lib/plurimath/formatter/numbers/fraction.rb
@@ -21,7 +21,7 @@ module Plurimath
 
           number = interpolate(format(options), fraction, :left)
           number = digit_count_format(int, fraction, number) if @digit_count
-          formatted_number = change_format(precision, number) if number
+          formatted_number = change_format(number) if number
           formatted_number ? decimal + formatted_number : ""
         end
 
@@ -32,7 +32,7 @@ module Plurimath
 
         protected
 
-        def change_format(precision, string)
+        def change_format(string)
           tokens = []
           tokens << string&.slice!(0, (group || string.length)) until string&.empty?
           tokens.compact.join(separator)
--- a/lib/plurimath/formatter/numbers/fraction.rb
+++ b/lib/plurimath/formatter/numbers/fraction.rb
@@ -45,12 +45,32 @@ module Plurimath
           @digit_count ||= int_length
           if int_length > @digit_count
             number_string = float.round(@digit_count - int.length)
-            @digit_count > int.length ? number_string.to_s("F").split(".").last : nil
+            numeric_digits(number_string) if @digit_count > int.length
           elsif int_length < @digit_count
-            number + "0" * (@digit_count - int_length)
+            number + "0" * (update_digit_count(number) - int_length)
           else
             number
           end
+        end
+
+        def numeric_digits(num_str)
+          float = num_str.to_s("F")
+          return float.split(".").last if float.length == @digit_count + 1
+          return unless @digit_count + 1 > float.length
+
+          float.split(".")[0] + ("0" * (@digit_count - float.length + 1))
+        end
+
+        def update_digit_count(number)
+          return @digit_count unless all_zeros?(number) == @precision
+
+          @digit_count - @precision + 1
+        end
+
+        def all_zeros?(number)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by suleman-uzair** [comment]
Hi @opoudjis

I have implemented support for the **`Significant`** option. I have added specs for some basic scenarios, could you please review and let me know if there are any edge cases or additional scenarios that need to be covered?

This is for when you’re back.

**Comment by opoudjis** [comment]
Cool, thank you! Will review tonight, after Metanorma release

**Review by opoudjis — APPROVED** [review]
Looks good, although I'd like to see tests under context "testing examples with de locale and significant option" for 0.1999 etc, as well as for numbers larger than zero.

**Review by opoudjis — CHANGES_REQUESTED** [review]
No, found problems:

* parameters  appear to persist between calls, even if the Plurimath object is reinitialised:
* significant applied to floats seems to be preserving all decimal digits in BigDecimals

so running isodoc, bin/rspec --format doc s*/i*/presentation_xml_maths_spec.rb

I'm getting:

````
Plurimath::NumberFormatter.new(de).localized_number(642121496772.6451564515, precision: 2, format: {:fraction_group_digits=>2, :fraction_group=>"'", :precision=>2, :locale=>:de, :significant=>10, :group=>"x", :group_digits=>3, :decimal=>","}) == 642x121x496x800

Plurimath::NumberFormatter.new(:fr).localized_number(30000, precision: 2, format: {:fraction_group_digits=>2, :fraction_group=>"'", :precision=>2, :locale=>:fr, :significant=>nil}) == 30'000,00

 Plurimath::NumberFormatter.new(:en).localized_number(31.0, precision: , format: {:fraction_group_digits=>2, :fraction_group=>"'", :notation=>:basic, :significant=>2}) == 31

Plurimath::NumberFormatter.new(:en).localized_number(327428.7432878432992, precision: , format: {:fraction_group_digits=>2, :fraction_group=>"'", :notation=>:basic, :significant=>19}) == 327x428,74'32'87'84'32'99'2 # it's remembering the :group setting from a few calls back

Plurimath::NumberFormatter.new(:en).localized_number(327428.7432878432992, precision: , format: {:fraction_group_digits=>2, :fraction_group=>"'", :significant=>3}) == 327x000

Plurimath::NumberFormatter.new(:en).localized_number(327428.7432878432992, precision: , format: {:fraction_group_digits=>2, :fraction_group=>"'", :significant=>9}) == 327x428.74'30'00'00'00'00'0 # that's not truncating the number of significant digits as expected

**Comment by opoudjis** [comment]
To replicate, please look at

https://github.com/metanorma/isodoc/pull/582

When you run

bin/rspec --format doc s*/i*/presentation_xml_maths_spec.rb

You will see that :group is unexpectedly persisting (setting it to nil prevents that, but I shouldn't have to on a new Plurimath object), and there is quite odd behaviour on significant when it includes both integer and decimal digits.

**Comment by suleman-uzair** [comment]
1. IsoDoc overrides the localisation of numbers in MathML with data-metanorma-numberformat attributes and default precision
2. IsoDoc overrides the localisation of numbers in MathML with data-metanorma-numberformat attributes and no default precision

@opoudjis I have fixed the configuration persistence issue, due to which the above-mentioned specs were failing.
Now you need to pass the `format` options hash each time you call the `localized_number()` function otherwise, the already persisted configuration will be used.
Let me know if any other changes are required.

**Comment by opoudjis** [comment]
Thank you, testing now

**Comment by opoudjis** [comment]
The persistence has been fixed. The odd behaviour of significant has not. So, with these called sequentially:

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :significant=>3}) == "327,000"
````

Correct. But immediately afterwards:

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :significant=>9}) == "327,428.74'30'00'00'00'00'0"
````

That's not correct: the significant value should be deleting trailing zeroes after the end of the significant digits if they are fractional: they are set to zero instead of truncated only in the integer part of the number. So this should be: "327,428.74'3"


````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>".", :notation=>:basic, :significant=>3}) == "327,000"
````

Again correct.

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:basic, :significant=>9}) == "327,42'8"
````

...?! This should be "327,428,74'3". Admittedly we now have an ambiguity with both group and decimal being comma, so I'll change the test, and that is an edge case; but I don't think the code should get confused by that. In any case,

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:basic, :significant=>9, :group=>"."}) == "327.428,74'30'00'00'00'00'0"
````

which goes back to the earlier failure to truncate trailing fractional zeroes.

Later on:

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:e, :significant=>9}) == "3,27'42'87'43'00'00'00'00'00e5"
````

Again, same failure to truncate trailing zeroes. 

Exponential notation is correct for small numbers, but badly wrong for very small or very large numbers, and that is independent of `significant`

````
Plurimath::NumberFormatter.new(:en).localized_number('1.1', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>4}) == "1,10'0e0"
````

Correct, but:

````
Plurimath::NumberFormatter.new(:en).localized_number('1000000000000000000000.0', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>1}) == "10 000 000 000 000 000 000 000e+21"
````

This should be just  "1e21".


````
Plurimath::NumberFormatter.new(:en).localized_number('10000000000000000000.0', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>2}) == "1,00'00'00'00'00'00'00'00'00'00e+19"
````

This should be just  "1.0e19".



````
Plurimath::NumberFormatter.new(:en).localized_number('0.0000000000000000001', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>2}) == "10 000 000 000 000 000 000e-20"
````

This should be just  "1.0e-19".

I'm getting inconsistent behaviour with the exponent plus: it is showing up even when I don't ask for it:

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>".", :notation=>:e, :significant=>3}) == "3.27'00'00'00'00'00'00'00'00e5"

Plurimath::NumberFormatter.new(:en).localized_number('3274287432878.432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>".", :notation=>:e, :significant=>3}) == "3.27'00'00'00'00'00'00'00'00e+12"
````

I suspect that if the exponent is greater than 9, the plus sign is being added. That is unwarranted: the plus sign must be added only if explicitly requested in the format parameters.

**Comment by opoudjis** [comment]
Scientific notation is also confused by significant:

````
Plurimath::NumberFormatter.new(:en).localized_number('1000000000000000000000.0', precision: 7, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:scientific}) == "1,00'00'00'0 × 10^+21"

Plurimath::NumberFormatter.new(:en).localized_number('1000000000000000000000.0', precision: 7, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:scientific, :significant=>1}) == "10 000 000 × 10^+21"
````

The latter of course should be just "1^+21"

**Comment by suleman-uzair** [comment]
> Admittedly we now have an ambiguity with both group and decimal being comma

We are [spliting](https://github.com/plurimath/plurimath/pull/268/files#diff-04d7674eddc1792c544c8e40cc6f25b965200a7996c4f61357468469860bd0daR23) numeric string based on the decimal symbol in post processing for significant digits feature. Additionally, having the same symbol for group/fraction_group and decimal will always be ambitious. Maybe we should return an error telling user to use different group/fraction_group and decimal symbol. Wdyt ?

```
Plurimath::NumberFormatter.new(:en).localized_number(‘0.0000000000000000001’, precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>“’”, :decimal=>“,”, :group=>” “, :notation=>:e, :significant=>2}) == “10 000 000 000 000 000 000e-20”
```
> This should be just “1.0e-19".

Your expected value is correct, But I see `sprintf` syntax we are using for notations creating some output issues for some very small numbers.
For example,
```
sprintf(“%.19e” %‘0.0000000000000000001’) => “9.9999999999999997525e-20"
sprintf(“%.19e” %‘0.000000000000000001’) => “1.0000000000000000715e-18"
```
A slight difference in the input values results in a notable change in the exponent value by a factor of 2.

@opoudjis I have fixed and added the specs for most of the scenarios you mentioned.

opoudjis:
So:

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :significant=>9}) == "327,428.74'30'00'00'00'00'0"
````

Fixed: "327,428.74'3"


````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:basic, :significant=>9, :group=>"."}) == "327.428,74'30'00'00'00'00'0"
````

Fixed: "327,428.74'3"

````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:e, :significant=>9}) == "3,27'42'87'43'00'00'00'00'00e5"
````

Fixed: "3,27'42'87'43e5" 

````
Plurimath::NumberFormatter.new(:en).localized_number('1000000000000000000000.0', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>1}) == "10 000 000 000 000 000 000 000e+21"
````

Fixed:  "1e21".

````
Plurimath::NumberFormatter.new(:en).localized_number('10000000000000000000.0', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>2}) == "1,00'00'00'00'00'00'00'00'00'00e+19"
````

Fixed: "1,0e19".

````
Plurimath::NumberFormatter.new(:en).localized_number('0.0000000000000000001', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :group=>" ", :notation=>:e, :significant=>2}) == "10 000 000 000 000 000 000e-20"
````

This should be just  "1,0e-19". It is now "10e-20". That is still a bug: the number before the exponent is meant to be between 0 < n < 10.

````
Plurimath::NumberFormatter.new(:en).localized_number('3274287432878.432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>".", :notation=>:e, :significant=>3}) == "3.27'00'00'00'00'00'00'00'00e+12"
````

Fixed: exponent_sign plus is showing up only if requested.

So there is one remaining bug to fix: "10e-20" should be "1,0e-19"

opoudjis:
> > Admittedly we now have an ambiguity with both group and decimal being comma
> 
> We are [spliting](https://github.com/plurimath/plurimath/pull/268/files#diff-04d7674eddc1792c544c8e40cc6f25b965200a7996c4f61357468469860bd0daR23) numeric string based on the decimal symbol in post processing for significant digits feature. Additionally, having the same symbol for group/fraction_group and decimal will always be ambitious. Maybe we should return an error telling user to use different group/fraction_group and decimal symbol. Wdyt ?
> 
> ```
> Plurimath::NumberFormatter.new(:en).localized_number(‘0.0000000000000000001’, precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>“’”, :decimal=>“,”, :group=>” “, :notation=>:e, :significant=>2}) == “10 000 000 000 000 000 000e-20”
> ```
> 
> > This should be just “1.0e-19".
> 
> Your expected value is correct, But I see `sprintf` syntax we are using for notations creating some output issues for some very small numbers. For example,
> 
> ```
> sprintf(“%.19e” %‘0.0000000000000000001’) => “9.9999999999999997525e-20"
> sprintf(“%.19e” %‘0.000000000000000001’) => “1.0000000000000000715e-18"
> ```
> 
> A slight difference in the input values results in a notable change in the exponent value by a factor of 2.
> 
> @opoudjis I have fixed and added the specs for most of the scenarios you mentioned.

So, about this:

Floating point issues are going to do things like this. And that is precisely why this gem must not be using Floating point arithmetic ever to deal with formatting: it must only use BigDecimal, since BigDecimal is intended to prevent such issues.

And because `sprintf(“%.19e”)` is intrinsically a floating point operation, that means that the numeric formatter must under no circumstances, ever at all, use sprintf. It is quite simply unfit for what we are doing.
You must only use https://ruby-doc.org/stdlib-3.1.0/libdoc/bigdecimal/rdoc/BigDecimal.html#method-i-to_s , and manipulate its output.

suleman-uzair:
> This should be “327,428,74'3". Admittedly we now have an ambiguity with both group and decimal being comma, so I’ll change the test, and that is an edge case;

I have created a separate issue [#270](https://github.com/plurimath/plurimath/issues/270) to fix the edge case you mentioned to address the specific scenario you mentioned. I will open a separate PR for this issue.

> So there is one remaining bug to fix: “10e-20” should be “1,0e-19". I note that when precision is set to 2, this gives the correct answer of “1,00e-19”.

@opoudjis I have fixed the remaining bug.

opoudjis:
You have indeed fixed the bug. 

But now there is a new bug.

````
Plurimath::NumberFormatter.new(:en).localized_number('642121496772645156.4515', precision: 7, format: {:fraction_group_digits=>4, :fraction_group=>"y", :precision=>7, :digitcount=>"10", :group=>"x", :group_digits=>3, :decimal=>",", :notation=>:scientific, :e=>"EE", :significant=>nil})
````

was correctly rounding up to `6,4212y150 × 10^17`. Now it is truncating instead of rounding: `6,4212y149 × 10^17`.

suleman-uzair:
@opoudjis It’s actually a typo and not a bug,
> Plurimath::NumberFormatter.new(:en).localized_number(‘642121496772645156.4515’, precision: 7, format: {:fraction_group_digits=>4, :fraction_group=>“y”, :precision=>7, :digitcount=>“10", :group=>“x”, :group_digits=>3, :decimal=>“,”, :notation=>:scientific, :e=>“EE”, :significant=>nil})

In your example, you are passing `digitcount` instead of `digit_count` and the value is also a string and not an integer; changing this fixes the output, which you can verify in the [specs](https://github.com/plurimath/plurimath/pull/268/commits/29ddd942b51333527c9080791e819a2b3539024e#diff-99bb7e6a5f1c520d1ed8f65c767a59420ebae04550ec191a4bd869f2eb7345ccR276).

opoudjis:
thank you!

suleman-uzair:
@opoudjis just released version `0.8.11`.

ronaldtse:
Thanks @suleman-uzair !

I had a discussion with @opoudjis and it seems that the number formatter only produces AsciiMath output. 

1. Being Plurimath we certainly should output the formatted number in all supported formats. 
2. Numbers in a formula, just like units, should already be formatted.

So we should implement the NumberFormatter to accept (parse) and generate the various formats. 

Can you help ticket this? This will also bring us closer to the ISO 80000 Math profile. Thanks!

Ron

PR #269, opened 2024-07-04
Updated Plurimath internal wrapper for symbols support

### LINKED ISSUE #217: Enable using backwards epsilon in AsciiMath (`\backepsilon` in LaTeX math)
`\backepsilon` symbol (϶, "such that" in math) is not available in AsciiMath. Is it possible to create a notation for it, to avoid using `latexmath` blocks just because of this symbol?

suleman-uzair:
This PR updates the **Plurimath** wrapper from `__{<symbol>}` to the following.
1. **UnicodeMath** and **Asciimath** wrapper => `"P{<symbol>}"`
2. **LaTeX** wrapper => `\text{P[symbol]}`

closes #217


ronaldtse:
@suleman-uzair this needs to be rebased. Thanks.

opoudjis (commented):
So... you only need to do the escaping in lib/plurimath/unicode_math/constants.rb, and nowhere else? Because you've removed it everywhere else?

I guess...

suleman-uzair:
> So... you only need to do the escaping in lib/plurimath/unicode_math/constants.rb, and nowhere else? Because you've removed it everywhere else?
> 
> I guess...

@opoudjis Yes, We need the escaping in _`lib/plurimath/unicode_math/constants.rb`_ for now.
The parsing rules for **UnicodeMath** are different from **AsciiMath** and **LateX**, which require **hexacode** and **English** letters as input.

ronaldtse (approved):
Thanks @suleman-uzair ! Let's add a blog post regarding compatibility wrappers.

PR #267, opened 2024-06-27
[WIP] Image generation for mathematical equations from CLI

suleman-uzair:
This PR supports image rendering from the CLI of the mathematical equations.

closes #205

PR #283, opened 2024-08-20
[WIP] Native wrapper around Lasem

### LINKED ISSUE #279: Ruby wrapper around Lasem for Plurimath image rendering
Apparently the `mathematical` library provides a wrapper around lasem. However, its lasem bindings are no longer maintained.

Maybe we have to reference them and do our own:
* https://github.com/gjtorikian/mathematical/tree/main/ext

_Originally posted by @ronaldtse in https://github.com/plurimath/plurimath/issues/267#issuecomment-2243067430_

suleman-uzair:
This PR incorporates C extension for Wrapping Lasem in **Plurimath** for image rendering.

closes #279


ronaldtse:
@suleman-uzair thanks, please move this to a new repository e.g. plurimath/lasem-ruby and a new gem `lasem`.

ronaldtse:
@suleman-uzair I've created the [lasem-ruby repository](https://github.com/plurimath/lasem-ruby).

suleman-uzair:
We can reference the [mathematical library](https://github.com/gjtorikian/mathematical/tree/main/ext) and then make changes on top of it in Plurimath([lasem-ruby repository](https://github.com/plurimath/lasem-ruby)). If we want to go in this direction, I can explore this further, i.e., making a C extension for Lasem.
@ronaldtse I want to understand our long-term goal here as well.
If we continue with this approach, it will not be compatible with plurimath-js as @hmdne mentioned earlier.
> Opal can change Ruby to JS, not C to JS.

If there are plans to incorporate image rendering in plurimath-js by using the same library in the future, we may need to identify a JS tool for Opal compatibility.
What do you think?

ronaldtse:
For the JS side, there is no need to render to SVG because MathJax does it well.

plurimath-js can act as a pre-processor when using MathJax or just directly to MathML for display, we do not need to generate PNGs.

We can adopt ideas and code from `mathematical` for `lasem-ruby`, but I do not want that to be mixed with Plurimath. It will require pre-compilation of gems on all various platforms and the workflows will be complicated.

i.e can you help combine #267 and this #283, then move the C/ext code to `lasem-ruby`?

suleman-uzair:
> For the JS side, there is no need to render to SVG because MathJax does it well.

@ronaldtse Sure, I'll continue with the **Lasem** integration and move the **C/ext** code `lasem-ruby` as you mentioned.

ronaldtse:
Closing in favor of new #448.

PR #291, opened 2024-09-19
List of classes supporting intent encoding

suleman-uzair:
This pull request includes a compiled list of **Plurimath** classes that support `intent` encoding for converting to **MathML**.


ronaldtse (commented):
Why not generate this using code?

suleman-uzair:
> Why not generate this using code?

@ronaldtse The intent list is not generated dynamically because it's limited to a few classes.

ronaldtse:
@suleman-uzair but I don't want to maintain a list. Let's generate this from code.

PR #304, opened 2024-10-04
Mathml conversion using LutaML-Model

### LINKED ISSUE #302: MathML parsing and transformation using lutaml-model
Update **MathML** parsing and transformation structure from using **Ox** (for parsing input) and **Parslet** (for transforming parsed input into a formula) to **Lutaml-Model**.

suleman-uzair:
This PR updates the parsing structure for **MathML** from using **Ox**(for parsing) and **Parslet** (for transformation).

close #302


lib/plurimath/math/formula/mrow.rb
```diff
@@ -0,0 +1,159 @@
+# frozen_string_literal: true
+
+module Plurimath
+  module Math
+    class Formula
+      class Mrow < Formula
```
ronaldtse, line 6: Is this `mrow` (and `mstyle`) a concept corresponding to MathML or is this to represent MathML? i.e. is this a generic concept? If it's a generic concept, should we just call it `row` and `style`?

lib/plurimath/math/function/menclose.rb
```diff
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
 require_relative "binary_function"
+require_relative "../../mathml/utility"
 
 module Plurimath
   module Math
     module Function
       class Menclose < BinaryFunction
```
ronaldtse, line 9: Similar to my comment above whether this is generic or MathML specific?

lib/plurimath/math/formula/mrow.rb
suleman-uzair: @ronaldtse, no, they are not generic and exist only for **MathML** parsing using **LutaML-Model**.

lib/plurimath/math/function/menclose.rb
suleman-uzair: This class itself is not generic but other conversions output are supported.

HassanAkbar (approved):

ronaldtse:
@suleman-uzair I have a question. For files like:
* lib/plurimath/math/function/fenced.rb

Is it possible to directly utilize lutaml-model to handle them?

Actually now we are talking about a `Plurimath::Math` model transformation into the `Mml` model which goes into MathML XML.

i.e.
```ruby
module Plurimath
  module Math
    module Function
      class TernaryFunction < Lutaml::Model::Serializable
        attribute :parameter_one, Math::Function::Core
        attribute :parameter_two, Math::Function::Core
        attribute :parameter_three, Math::Function::Core
        attribute :hide_function_name, :boolean

        def initialize(*args)
          super
          Utility.validate_left_right(variables.map { |var| get(var) })
        end
# ...
end

module Plurimath
  module Math
    module Function
      class Fenced < TernaryFunction
        include Mathml::Utility

        attribute :separators, Math::Symbol, validate: -> { |value| validate_symbols(value) }
        attribute :content, Math::Function::Core
        attribute :open, Math::Symbol, validate: -> { |value| validate_symbols(value) }
        attribute :close, Math::Symbol, validate: -> { |value| validate_symbols(value) }
        # ...
end
```

ronaldtse (approved):

ronaldtse:
@suleman-uzair since this is done let's merge it first. Thank you for the good work!

Gemfile
```diff
@@ -13,3 +13,4 @@ gem 'oga'
 gem 'ox'
 gem "unitsml"
 gem "debug"
+gem "mml", git: "https://github.com/plurimath/mml", branch: "features/mathml_parsing"
```
ronaldtse, line 16: We need to remove the git address, I'm sure you will...

suleman-uzair: > We need to remove the git address

@ronaldtse, Yes, I’ll do that but we’ll need to release the `mml` gem’s version before that.
Can you review/approve the plurimath/mml#2 first, so I can merge and release the version first?

PR #305, opened 2024-10-04
`number_sign` option in *NumberFormatting*

### LINKED ISSUE #296: Add number_sign option for number formatting
From https://github.com/metanorma/metanorma-standoc/issues/930

We need a `number_sign` option for number formatting, with behaviour analogous to `exponent_sign` (for consistency). If it is present, and the number is positive, then put the number_sign symbol in front of it. As with `exponent_sign`, a value of `plus` is interpreted as `+`.

So 

````
Plurimath::NumberFormatter.new()
  .localized_number(-5, format: { number_sign: :plus }) = "-5"
Plurimath::NumberFormatter.new()
  .localized_number(5, format: { number_sign: :plus }) = "+5"
Plurimath::NumberFormatter.new()
  .localized_number(0, format: { number_sign: :plus }) = "0"
Plurimath::NumberFormatter.new()
  .localized_number(50, format: { number_sign: :plus, notation: :e }) = "+5e1"
Plurimath::NumberFormatter.new()
  .localized_number(50, format: 
  { number_sign: :plus, exponent_sign: :plus, notation: :e }) = "+5e+1"
````

### PR DESCRIPTION
**Pull request by suleman-uzair**
This PR supports the `number_sign` option in the `format` argument in the **NumberFormatter** feature.
Additionally, it updates the `Standard` class options list in the method `set_default_options` and in the constant `DEFAULT_OPTIONS`.

close #296

### CODE REVIEW

**Comment by suleman-uzair** [comment]
@opoudjis, The current implementation of **NumberFormatter** is that negative numbers as input will be converted to positive in all cases, following the **twitter_cldr** gem's functionality(you can see the code [here](https://github.com/twitter/twitter-cldr-rb/blob/master/lib/twitter_cldr/formatters/numbers/number_formatter.rb#L74)).
Currently `number_sign:` supports only value `:plus`.

Also, To be on the same page, the following are the expected behaviors of the given examples in issue #296.


```ruby
Plurimath::NumberFormatter.new()
  .localized_number("-5", format: { number_sign: :plus }) = "5" # never "-5" in any case
Plurimath::NumberFormatter.new()
  .localized_number("5", format: { number_sign: :plus }) = "+5" # successfully inserts plus sign
Plurimath::NumberFormatter.new()
  .localized_number("0", format: { number_sign: :plus }) = "0" # won't insert plus sign
Plurimath::NumberFormatter.new()
  .localized_number("50", format: { number_sign: :plus, notation: :e }) = "+5.0e1" # inserts plus sign
Plurimath::NumberFormatter.new()
  .localized_number(50, format: 
  { number_sign: :plus, exponent_sign: :plus, notation: :e }) = "+5.0e+1" # inserts plus sign
```

Let me know if this will cover all cases or if we need to extend support here in some way?

opoudjis:
````
Plurimath::NumberFormatter.new()
  .localized_number("-5", format: { number_sign: :plus }) = "5" # never "-5" in any case
````

Um... I can't use that. Moreover, that's not what that twitter-cldr-rb code is doing: that code is extracting the (negative) sign as the initial token, and it is sticking it back on after processing: numbers are being processed as absolute, but the sign is being kept apart.  And twitter-cldr-rb differentiates minus_sign and plus_sign; you can see that they are localised as two separate things in https://github.com/twitter/twitter-cldr-rb/blob/29758ac8ae284cd98becee90dd698bf1fa0e252f/lib/twitter_cldr/formatters/numbers/helpers/integer.rb#L46


````
-4.3.localize(:de).to_s
````

is executed in twitter_cldr_rb as:

````
prefix, suffix, integer_format, fraction_format = *partition_tokens(tokens)
# prefix == "-", suffix = ""; number = -4.3
number = truncate_number(number, integer_format.format.length)
# number = -4.3
int, fraction = parse_number(number, options)
# int = "4", fraction = "3", sign is indeed discarded
result =  integer_format.apply(int, options)
result << fraction_format.apply(fraction, options) if fraction
# result = "4,3"
number_system.transliterate(
  "#{prefix.to_s}#{result}#{suffix.to_s}"
)
# so the number goes back to "-4,3"

In fact, twitter-cldr-rb does something more underhand: `tokens` is generated as `-#,##0.###` if the number is negative, and `#,##0.###` if the number is positive. So the sign is sidestepped in generation anyway.

All this is to say, the behaviour implemented is not the behaviour desired, and it cannot be justified by the way twitter_cldr_rb is implemented, which removes sign only to put it back in.

````
Plurimath::NumberFormatter.new()
  .localized_number("-5", format: { number_sign: :plus }) = "-5" # change: number_sign does not override minus sign, just as it doesn't in exponent_sign
````

opoudjis (changes requested):
See comment

suleman-uzair:
@opoudjis, I have fixed the missing minus sign behavior for negative numbers.
Can you please review and let me know if the PR meets your requirements now?

opoudjis (approved):
Thank you, behaves as expected:

````
<p id="_8c6e3796-0a99-0d4e-47ca-9c936df41ab5">
<stem type="MathML">
<math xmlns="http://www.w3.org/1998/Math/MathML">
<mn data-metanorma-numberformat="number_sign='plus'">0.4e1</mn>
</math>
</stem>  
<stem type="MathML">
<math xmlns="http://www.w3.org/1998/Math/MathML">
<mn data-metanorma-numberformat="number_sign='plus'">-0.4e1</mn>
</math>
</stem>
</p>
````

ends up rendering as

````
<p id="_8c6e3796-0a99-0d4e-47ca-9c936df41ab5">+4  -4</p>
````

suleman-uzair:
@opoudjis , Just released version `0.8.17`.

ronaldtse:
@suleman-uzair please help update documentation and the blog post, thanks.

PR #329, opened 2025-01-27
Updated `mml` configuration and added `display` methods

suleman-uzair:
This PR moves the `Mml` configuration from the method to the file's top level.
Add the `display` attribute's methods (added in `mml`) in `formula.rb`.

related => plurimath/mml#4

### DIFF
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -585,6 +585,16 @@ module Plurimath
         update_temp_order(value, "mglyph")
       end
 
+      def display
+        @display
+      end
+
+      def display=(value)
+        return unless value
+
+        @display = value
+      end
+
       protected
 
       def update_temp_order(value, order_name)
--- a/lib/plurimath/math/formula.rb
+++ b/lib/plurimath/math/formula.rb
@@ -7,7 +7,7 @@ module Plurimath
     class Formula < Core
       include Mathml::Utility
 
-      attr_accessor :value, :left_right_wrapper, :displaystyle, :input_string, :unitsml, :unitsml_xml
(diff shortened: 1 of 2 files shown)


lib/plurimath/math/formula.rb
```diff
@@ -585,6 +585,16 @@ def mglyph_value=(value)
         update_temp_order(value, "mglyph")
       end
 
+      def display
+        @display
+      end
+
+      def display=(value)
+        return unless value
```
HassanAkbar, line 593: This will not allow to unset the `@display` value once it is set. May be we need an accessor here?

HassanAkbar (approved):

lib/plurimath/math/formula.rb
```diff
@@ -7,7 +7,7 @@ module Math
     class Formula < Core
       include Mathml::Utility
 
-      attr_accessor :value, :left_right_wrapper, :displaystyle, :input_string, :unitsml, :unitsml_xml
+      attr_accessor :value, :left_right_wrapper, :displaystyle, :input_string, :unitsml, :unitsml_xml, :display
```
ronaldtse, line 10: Can we use lutaml-model here (in a new PR)?

lib/plurimath/mathml/parser.rb
```diff
@@ -1,66 +1,60 @@
...
     class Parser
       attr_accessor :text
 
       def initialize(text)
-        mml_config
         @text = text
       end
 
       def parse
         namespace_exist = text.split(">").first.include?(" xmlns=")
```
ronaldtse, line 55: Also wish to get rid of native XML string matching stuff like that, should migrate to lutaml-model.

ronaldtse (approved):

PR #338, opened 2025-02-10
Unitsml-Ruby added as dependency

### LINKED ISSUE #335: (URGENT) Issue with "unitsml" gem dependency
This particular line requires the "unitsml" gem, but the gem is not in the gemspec! That means that when I require "plurimath" from another gem (https://github.com/metanorma/coradoc/issues/144), a library not found error is raised.

https://github.com/plurimath/plurimath/blob/c1e816410bfd3ebbc86c9a09e2f48db7b249e895/lib/plurimath/unitsml.rb#L3

This forces the developer to require both gems, "plurimath" and "unitsml".

We need to break this circular dependency.

In Coradoc we are forced to add the "unitsml" gem as a dependency:
* https://github.com/metanorma/coradoc/pull/148/files#diff-630b499198738f7d1c2bdae57acdf74f13acdf5ff4d97f164d0530d34dbc1e13R33

suleman-uzair:
This PR adds the `Unitsml-Ruby` gem as a `dependency` in the `gemspec` file.

closes #335

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,5 +11,4 @@ gem 'equivalent-xml'
 gem 'opal-rspec', "~> 1.1.0a"
 gem 'oga'
 gem 'ox'
-gem "unitsml"
 gem "debug"
--- a/plurimath.gemspec
+++ b/plurimath.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'mml'
   spec.add_dependency 'thor'
   spec.add_dependency 'parslet'
+  spec.add_dependency 'unitsml'
   spec.add_dependency 'bigdecimal'
   spec.add_dependency 'lutaml-model'
 end


suleman-uzair:
@ronaldtse, This PR and [unitsml/unitsml-ruby#29](https://github.com/unitsml/unitsml-ruby/issues/29) reverses the dependency with a little change.
- Changes in **Plurimath**:
   a. Before: Users had to add both **Plurimath** and **Unitsml-Ruby** to their Gemfile.
   b. Now: **Unitsml-Ruby** is installed automatically, which fixes issue #335.

- Changes in **Unitsml-Ruby**:
   a. Before: **Plurimath** was a required dependency.
   b. Now: **Plurimath** is only required when using the `to_plurimath` function.

Summary: **Plurimath** depends on the **Unitsml-Ruby**, and **Unitsml-Ruby** will only require/need **Plurimath** when `to_plurimath` is called.

Please let me know if you have any other approach in mind.

ronaldtse (approved):

PR #345, opened 2025-03-16
Fix: `Function::Unitsml` instances duplication

### LINKED ISSUE #343: Error: Maths: Malformed MathML: [plurimath] Error: Failed to parse the following formula with type `asciimath`.
The source document: https://github.com/metanorma/bipm-si-brochure/blob/main/sources/si-brochure/3.01/si-brochure-en.adoc

`bundle exec metanorma -t bipm -x presentation si-brochure-en.adoc` fails with the errors:

```
---- FORMULA BEGIN ----
"unitsml(kg^(-1)*m^(-3)*s^4*A^2)"
---- FORMULA END ----

<stem block="false" type="MathML">"unitsml(kg^(-1)*m^(-3)*s^4*A^2)"</stem>
Maths: Malformed MathML: [plurimath] Error: Failed to parse the following formula with type `asciimath`.
[plurimath] Please first manually validate the formula.
[plurimath] If this is a bug, please report the formula at our issue tracker at:
[plurimath] https://github.com/plurimath/plurimath/issues
```

suleman-uzair:
This PR fixes the duplication of `Plurimath::Math::Function::Unitsml` instances.

closes #343

### DIFF
--- a/lib/plurimath/math/core.rb
+++ b/lib/plurimath/math/core.rb
@@ -213,7 +213,12 @@ module Plurimath
       end
 
       def cloned_objects
-        object = self.class.new rescue self.class.new(nil)
+        object = begin
+                   self.class.new
+                 rescue
+                   content = text if class_name == "unitsml"
+                   self.class.new(content)
+                 end
         variables.each { |var| object.set(var, variable_value(get(var))) }
         object
       end
--- a/lib/plurimath/math/core.rb
+++ b/lib/plurimath/math/core.rb
@@ -213,12 +213,7 @@ module Plurimath
       end
 
       def cloned_objects
-        object = begin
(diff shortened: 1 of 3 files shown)


lib/plurimath/math/core.rb
```diff
@@ -213,7 +213,12 @@ def linebreak?
       end
 
       def cloned_objects
-        object = self.class.new rescue self.class.new(nil)
+        object = begin
```
ronaldtse, line 216: Using rescue triggers a lot of overhead internally. Why not just do a const check and instantiate?

ronaldtse (changes requested):

ronaldtse:
Thank you @suleman-uzair !

PR #364, opened 2025-04-17
Update: Significant digits

### LINKED ISSUE #363: (URGENT) Number formatter still generating wrong values
```yaml
      - geometrical-altitude: 0
        r-1: 0.0059
        e-1: 0.0082
        t-1: -62
        r-5: 0.0087
        e-5: 0.0141
        t-5: -58
        r-10: 0.0129
        e-10: 0.0209
        t-10: -55
        r-20: 0.0189
        e-20: 0.0307
        t-20: -52
```

The values are generating using this:
```
| {% if value.geometrical-altitude %} {{ value.geometrical-altitude }} {% endif %}
| {% if value.r-1 %}number:{{ value.r-1 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.e-1 %}number:{{ value.e-1 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.t-1 %}number:{{ value.t-1 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.r-5 %}number:{{ value.r-5 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.e-5 %}number:{{ value.e-5 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.t-5 %}number:{{ value.t-5 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.r-10 %}number:{{ value.r-10 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.e-10 %}number:{{ value.e-10 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.t-10 %}number:{{ value.t-10 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.r-20 %}number:{{ value.r-20 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.e-20 %}number:{{ value.e-20 }}[profile=iso80000]{% else %} - {% endif %}
| {% if value.t-20 %}number:{{ value.t-20 }}[profile=iso80000]{% else %} - {% endif %}
```

And the `iso80000` profile is:
```asciidoc
:number-presentation-profile-iso80000: group_digits=3,fraction_group_digits=3,decimal=",",group=&#x2009;,fraction_group=&#x2009;,notation=general,large_notation=scientific
```

Notice that in this table, all decimal values are rendered with extra zeros at the end:

* (0, 1%, r) 0.0059 => "0,005900"
* (1, 1%, r) 0.0442 => "0,04420"
* (0, 1% e') 0.0082 => "0,008200"

Image

Non-zero decimal numbers that have a '0' as the first position at the decimal portion also sees the same problem:

```yaml
      - geometrical-altitude: 6
        r-1: 8.8
        e-1: 6.56
        t-1: 1
        r-5: 8.18
        e-5: 6.09
        t-5: 0
        r-10: 7.59
        e-10: 5.67
        t-10: -1
        r-20: 7.05
        e-20: 5.27
        t-20: -2
```

* 6.09 => "6,090"
* 7.05 => "7,050"

Image

suleman-uzair:
This PR updates the behaviour of `significant` options passed inside `format` hash for number formatting.

#### Previous behaviour:
```ruby
> Plurimath::NumberFormatter.new(:en).localized_number("0.0082", format: { significant: 4 })
=> "0.008200" # adding significant zeros to meet the significant number passed.
```
#### New behaviour:
```ruby
> Plurimath::NumberFormatter.new(:en).localized_number("0.0082", format: { significant: 4 })
=> "0.0082" # If the value of the `significant` option exceeds the given number, then truncate the number; otherwise, do nothing.
```

closes #363

### DIFF
--- a/lib/plurimath/formatter/numbers/significant.rb
+++ b/lib/plurimath/formatter/numbers/significant.rb
@@ -33,9 +33,9 @@ module Plurimath
           frac_part = false
           new_chars = []
           sig_count = significant
-          chars.each.with_index do |char, ind|
-            frac_part = frac_part || char == decimal
-            sig_num = sig_num || char.match?(/[1-9]/)
+          chars.each_with_index do |char, ind|
+            frac_part ||= char == decimal
+            sig_num ||= char.match?(/[1-9]/)
             break if sig_count.zero?
 
             new_chars << char
@@ -47,7 +47,6 @@ module Plurimath
 
           if sig_count > 0
             new_chars << decimal unless frac_part
-            new_chars << ("0" * sig_count)
           else
             remain_chars = count_chars(chars, frac_part) - significant
             if remain_chars > 0
@@ -66,7 +65,7 @@ module Plurimath
 
           frac_part = false if chars[arr_len] == decimal
           prev_ten  = false
-          array.reverse!.each.with_index do |char, ind|
+          array.reverse!.each_with_index do |char, ind|
             if char == decimal
               array[ind] = ""
               frac_part  = false
--- a/spec/plurimath/number_formatter_spec.rb
+++ b/spec/plurimath/number_formatter_spec.rb
@@ -160,7 +160,7 @@ RSpec.describe Plurimath::NumberFormatter do
         output_string = formatter.localized_number(number, format: { digit_count: 6, notation: :scientific, group_digits: 3 })
         expect(output_string).to eql("1,00000 × 10^-3")
         output_string = formatter.localized_number(number, format: { significant: 3, group_digits: 3 })
-        expect(output_string).to eql("0,00100")
+        expect(output_string).to eql("0,001")
         output_string = formatter.localized_number(number, format: { significant: 3, notation: :e, group_digits: 3 })
         expect(output_string).to eql("1,00e-3")
         output_string = formatter.localized_number(number, format: { significant: 3, notation: :scientific, group_digits: 3 })


ronaldtse:
I am working with 2533 and 5878.

This thread is mixing up the input number and output format needs. 

The input number defines the number of significant digits and also the actual numbers. These cannot be changed. This provides the semantics. Mind you, we do not support "specification of significant digits" today. So 0.87 is 0.87, not 0.870000. 

The output string is the "formatted number". This string needs to adhere to the options I provide to the formatter. When I specify significant digits, it should not exceed what my original number contains. I.e. 0.87 cannot be formatted as 0.87000 in any circumstance, whether when I request for 6 or 100 digits, it is just 0.87. 

When my input number has more significant digits that the output format allows, eg 0.87123456 with 6 significant digits means 0.871234. 

When I have a significant digit setting and an exponent setting, the digits become 8.71234 x 10^-1. 

The complication here comes when I have both significant digits and digit count, and the digit count cuts off the significant digits. Normally when I have digits count, 0.012345 in 3 digit count is 0.01, but 3 significant digits is 0.123. When I have 1234.56789, digit count of 3 would be...??

Maybe digits count is just the wrong thing to have.

opoudjis:
>  When I specify significant digits, it should not exceed what my original number contains. I.e. 0.87 cannot be formatted as 0.87000 in any circumstance, whether when I request for 6 or 100 digits, it is just 0.87.

Given what can happen to numbers in preprocessing, including dropping trailing zeroes (which happens by default in computational types), you cannot guarantee "what your original number contains" is accurate. In any case, as I already said, I count digits prior to populating the formatting options, to prevent that—but only if significant numbers have not been specified already.

````
 def explicit_number_formatter_signif(num, symbols)
      signif = symbols[:significant]
      (symbols.keys & %i(precision digit_count)).empty? and
        signif ||= num_totaldigits(num.text)
      signif
    end
````

If you have both significant digits and digit count, clearly the digit count overrides the significant digits. The point of digit count is uniform presentation, no matter what else is going on. Having both significant digits and digit count and expecting them not to conflict is naive. It may well be preferable to document that you cannot have both in a number, and discard one of the two if you do get them.

I continue to disagree with the original premiss. If I passed 4 significant digits and 0.0087, I should be getting 4 significant digits. It is the user's fault for providing input that is contradictory, and I am more prepared to assume that the number was faulty than that the significant figures were faulty, because it is so easy to drop trailing zeroes.

ronaldtse:
The problem here is that the input number itself does not tell how many significant digits it has. 

> If I requested 4 significant digits and 0.0087, I should be getting 4 significant digits

This is very strange. The number only contains 2 significant digits. We cannot make up its accuracy by adding 2 more significant digits?

ronaldtse:
When given a number as input:
* significant digits can be set, then we use the specified number.
* significant digits not set, we have to auto-detect (after the string of consecutive 0's counting from the left, count the characters until the end at the right). e.g. `0.0007` has 1 significant digit, `0.0007000` has 4 significant digits.

When this number is output:
* significant digits to be shown must be `max(original number significant digits, formatted significant digits)`

ronaldtse (approved):

ronaldtse:
We need to update documentation too.

PR #366, opened 2025-04-19
Memoization and grammar optimizations

alex-sc:
- [Formula memoization](https://github.com/plurimath/plurimath/commit/c0729fce77a50286165528977ad5506a82756609)
- [Optimize asciimath grammar for the huge Constants.precompile_constant…](https://github.com/plurimath/plurimath/commit/8551fd0838981a5fd661abc6abf00c0eb01f8fe1)
- [Latex grammar optimization](https://github.com/plurimath/plurimath/commit/3888adb5a479736f1aa2c46bf68d537456e831ca)
@alex-sc

### DIFF
--- a/lib/plurimath/asciimath/parse.rb
+++ b/lib/plurimath/asciimath/parse.rb
@@ -78,7 +78,7 @@ module Plurimath
         sub_sup_classes |
           binary_classes |
           ternary_classes |
-          hash_to_expression(Constants.precompile_constants) |
+          hash_to_expression_grouped(Constants.precompile_constants) |
           (match(/[0-9]/).as(:number) >> str(",").as(:comma)).repeat(1).as(:comma_separated) |
           quoted_text |
           (str("d").as(:d) >> str("x").as(:x)).as(:intermediate_exp) |
@@ -187,6 +187,23 @@ module Plurimath
         end
       end
 
+      def hash_to_expression_grouped(arr)
+        @@expression ||= begin
+                           grouped = Hash.new { |h, k| h[k] = [] }
+                           arr.each do |expr_string|
+                             str, sym = expr_string
+                             grouped[sym] << str
+                           end
+                           grouped.reduce("") do |expression, expr_group|
+                             if expression.is_a?(String)
+                               dynamic_parser_rules_group(expr_group)
+                             else
+                               expression | dynamic_parser_rules_group(expr_group)
+                             end
+                           end
+                          end
+      end
+
       def dynamic_parser_rules(expr)
         first_value = str(expr.first.to_s)
         case expr.last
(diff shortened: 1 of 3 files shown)


alex-sc:
There changes make `bundle exec rake` work 2-2.2x faster

lib/plurimath/utility.rb
```diff
@@ -457,8 +458,11 @@ def string_to_html_entity(string)
       end
 
       def html_entity_to_unicode(string)
-        entities = HTMLEntities.new
-        entities.decode(string)
+        htmlEntities().decode(string)
```
ronaldtse, line 461: Strange to use `()` here. Is it needed? We can also make `Utility` a Singleton instead of having to access Thread variables.

alex-sc: @ronaldtse 
Right `()` is not needed.

I'm not sure if `HTMLEntities`/`encoder` is thread-safe

ronaldtse: Doesn't matter if it's thread safe, I think a Singleton only exists within the Thread...?

ronaldtse: Maybe we can use RactorLocalSingleton instead which was added here:
- https://github.com/ruby/singleton/pull/4

A Ractor seems to be implemented as a Thread.

alex-sc: Thanks, I will take a look in the evening

I propose to drop this particular optimization for now since it's a minor one and there many similar usages of `HTMLEntities`, `HTMLEntities::encoder`, `HTMLEntities::decoder` across different gems. There's another usage of `HTMLEntities` in `string_to_html_entity` in the same class

Let's remove this change and create a separate PR for `HTMLEntities`/RactorLocalSingleton. Do you agree @ronaldtse?

alex-sc: @ronaldtse I've removed the `HTMLEntities` change for now. Let's continue...

ronaldtse:
@suleman-uzair can you please help review this (and clean up caching methods/locations since you know the code better) and ensure that it works at Opal/Plurimath-JS? Thanks!

lib/plurimath/asciimath/parse.rb
```diff
@@ -179,21 +179,30 @@ def read_text
...
 
-      def hash_to_expression(arr)
-        type = arr.first.class
-        @@expression ||= arr.reduce do |expression, expr_string|
-          expression = dynamic_parser_rules(expression) if expression.is_a?(type)
-          expression | dynamic_parser_rules(expr_string)
+      def hash_to_expression_grouped(arr)
+        @@expression ||= begin
+          grouped = Hash.new { |h, k| h[k] = [] }
+          arr.each { |value, kind| grouped[kind] << value }
```
suleman-uzair: This re-groups the flat hash from `precompile_constants` back into per-kind arrays, undoing the merge that `Constants.precompile_constants` does. This can also be resolved by having `precompile_constants` return the data already grouped, with each group sorted by string length (longest first) to preserve correct Parslet match priority (e.g., `ddots` must match before `ddot`).

lib/plurimath/math.rb
```diff
@@ -38,14 +38,19 @@ module Math
       asciimath: Asciimath,
     }.freeze
 
+    # TODO: make optional/weak
+    FORMULA_CACHE = Hash.new { |h, k| h[k] = {} }
+
     def parse(text, type)
       raise InvalidTypeError.new unless valid_type?(type)
 
+      return FORMULA_CACHE[type][text] if FORMULA_CACHE[type].key?(text)
```
suleman-uzair: The cached **Formula** object is returned by reference. Since **Formula** is mutable, any downstream code that transforms or modifies the parsed formula could silently corrupt the cache:
```ruby
f1 = Plurimath::Math.parse("a^2 + b^2", :latex)
f1.value.clear
f2 = Plurimath::Math.parse("a^2 + b^2", :latex)
f2.value # => [] — corrupted
```
This can be fixed by returning a deep copy: `FORMULA_CACHE[type][text].cloned_objects` as `Core#cloned_objects` already handles recursive deep cloning of the entire Formula.

lib/plurimath/math.rb
```diff
@@ -38,14 +38,19 @@ module Math
       asciimath: Asciimath,
     }.freeze
 
+    # TODO: make optional/weak
+    FORMULA_CACHE = Hash.new { |h, k| h[k] = {} }
```
suleman-uzair, line 41: **FORMULA_CACHE** grows unboundedly in long-running processes (as @jcasimir also noted). Consider adding an LRU cap, a `clear_cache!` method, or making caching opt-in (e.g., `cache: true`, default `false`)

suleman-uzair (changes requested):
@ronaldtse I've already updated the PR with some DRY changes to the `asciimath` and `latex` parsers. I'll implement the following suggested changes.
Let me know if you have any further comments.

suleman-uzair:
Hi @ronaldtse,
The review changes for this PR have been finalized. Here is a brief summary of what was addressed:

* Refactored repeated reduce patterns in the `asciimath` and `latex` parsers for readability
* `precompile_constants` now returns a grouped hash, sorted by length per group
* Formula cache is enabled by default, with support to disable it via `cache: false` or clear it via `clear_cache!`
* Cache entries are protected via deep cloning on store and retrieval to prevent mutation corruption, with a minor performance overhead as an accepted trade-off for safety
* Double-struck symbols (**ZZ**, **RR**, **QQ**, **NN**, **CC**) updated to use standard representations across all output formats

Please let me know if you have any comments, suggestions, or feedback.

PR #406, opened 2026-03-25
fix: map combining left right arrow above (U+20E1) for OMML accent parsing

jcasimir:
## Summary

Adds `"&#x20e1;": "vec"` to UNICODE_SYMBOLS so the combining left right arrow above character (U+20E1) resolves to `Math::Function::Vec` during OMML accent parsing.

This character is used in Microsoft Word for geometry line notation (e.g., line $\overleftrightarrow{AB}$), as opposed to the combining overline (U+0305) used for segments. Same fix pattern as #405.

**Note:** `Vec` produces a right arrow (`→`) rather than a left-right arrow (`↔`). This is the closest existing function class — adding a dedicated `Overleftrightarrow` function would be more precise but is a larger change.

Fixes 38 additional EngageNY curriculum OMML parse failures, all involving geometry line notation.

## Test plan

- [x] All 38 U+20E1 accent expressions from EngageNY now parse successfully
- [x] OMML test suite passes (272 examples, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ronaldtse:
Thank you @jcasimir for this PR! All the changes here have been incorporated into `main` through the following commits:

- **30562a12** — fix: default to parentheses when OMML m:d omits begChr/endChr
- **9f8f4af7** — fix: handle OMML accent elements with val attributes (overbar/overline)
- **de94c51d** — fix: map combining left right arrow above (U+20E1) to vec
- **f332a070** — fix: OMML fixes, rubocop cleanup, and performance optimization

One improvement over the original U+20E1 mapping: instead of mapping to `vec` (which produces a right arrow →), it now maps to a dedicated `Overleftrightarrow` function (which produces a left-right arrow ↔), better matching the semantics of the combining left-right arrow above character.

Also included: the zero-in-scientific-notation fix, fixture 109 update, rubocop cleanup, and a performance optimization that makes OMML generation 2–11x faster.

Closing as superseded — thanks again!

PR #412, opened 2026-04-01
Refactor MathML Parsing, Translation, and Opal Loading

suleman-uzair:
This PR refactors the MathML parsing and translation flow.

- updates the `Mathml::Translator`-based conversion flow
- moves MathML-specific helper logic into `lib/plurimath/mathml/formula_transformation.rb`
- keeps translation entry points in `Translator` and reduces inline helper clutter
- normalizes MathML token handling for AST conversion
- updates affected MathML specs
- refines Opal bootstrap/loading support


suleman-uzair:
@ronaldtse, I have also been working on updating the MathML parsing/conversion flow, and it seems like we may be approaching it from two slightly different directions.

1. **Translator-based flow**  
   This keeps the conversion logic centralized in `lib/plurimath/mathml/translator.rb`, where MathML nodes are mapped to Plurimath instances in one place.  
   Pros:  
   - easier to trace the full conversion flow in one file  
   - simpler to compare against previous behavior during migration  

   Cons:
    - may drift from the `Mml` model/context structure over time  

2. **Registered model/context flow**  
   This uses the registered MathML model classes in `lib/plurimath/mathml/models/**`, with each model handling its own `to_plurimath`, and `parser.rb` relying on `Mml.parse(...).to_plurimath`.  
   Possible pros:  
   - aligns more closely with the `Mml` context/model structure  
   - keeps node-specific behavior with the corresponding model  

   Possible cons:  
   - migration is a bit harder to validate because behavior is spread across many model files  

My current understanding is that the registered model/context flow may align better with the long-term `Plurimath` + `Mml` structure, but I may be missing some context here.

Could you please let me know whether I should continue with the registered model/context flow or the Translator flow in this PR?

ronaldtse:
> may drift from the Mml model/context structure over time

Actually I am expecting this to happen, because in the near future we will want to handle content math models like OpenMath which will have a different model compared to MathML presentation. 

That's why the model Translator was added. 

In as to the method of how we register models, the new lutaml-model now supports a hierarchical register (register links to another register for types it cannot resolve), and the model registration mechanism is also used for complex versioning like in the lutaml/XMI gem. 

Could you have a look to see if we can migrate the Configuration object to align with those practices, or improve the default way of registration in lutaml-model? Ideally we want to keep only best practices and promote them across all work. 

Thanks!

ronaldtse:
@suleman-uzair are we ready to merge this? Thanks.

suleman-uzair:
> @suleman-uzair are we ready to merge this? Thanks.

@ronaldtse Yes, this PR is ready along with the following PRs:
* unitsml/unitsdb-ruby#25
* unitsml/unitsml-ruby#61

`metanorma-cli` doesn't support lutaml-model `0.8.0` yet, so I'm unable to verify these changes there.

PR #418, opened 2026-04-27
Fix significant digit formatting for base conversion

### LINKED ISSUE #417: Base conversion generating incorrect default precision count for fractional digits
## Summary

As noted in https://github.com/plurimath/plurimath/issues/416, isodoc in Metanorma works out how many significant digits are needed for a number, and passes that on to Plurimath. When there is a base conversion resulting in a smaller number of significant digits, that too is passed on, as implemented in metanorma/isodoc#788. But the precision is left to null by default, to be calculated by Plurimath.

Over to Claude:

When formatting a number in a non-decimal base with `:significant` set, Plurimath produces too few fractional digits because **decimal fractional precision is reused as target-base fractional precision**. For example `0.12325e3` (= 123.25) with `{base: 16, significant: 5}` produces `0x7b.40` (4 sig digits) instead of `0x7b.400` (5 sig digits).

This was reported downstream at [metanorma/isodoc#788](https://github.com/metanorma/isodoc/issues/788). isodoc has landed a workaround that computes target-base precision externally and passes it as `precision:`, but the underlying behavior should be fixed in Plurimath so callers don't have to.

## Reproducer

```ruby
require "plurimath"

result = Plurimath::NumberFormatter.new("en")
  .localized_number("0.12325e3", format: { base: 16, significant: 5 })

puts result        # => "0x7b.40"   (actual)
# expected        => "0x7b.400" with significant: 5

# Same value rendered with explicit precision works:
result_workaround = Plurimath::NumberFormatter.new("en")
  .localized_number("0.12325e3", precision: 3,
                                 format: { base: 16, significant: 5 })

puts result_workaround   # => "0x7b.400"
```

## Root cause pointers

Two pieces of code use precision with **mismatched semantics** when `:base` is non-decimal:

1. **`Plurimath::Formatter::NumberFormatter#parse_number` / `#precision_from`** ([lib/plurimath/formatter/number_formatter.rb:83-100](https://github.com/plurimath/plurimath/blob/master/lib/plurimath/formatter/number_formatter.rb#L83-L100)) compute precision from `BigDecimal#to_s("F")` — i.e. **count of decimal fractional digits** of the parsed numeric value, independent of `:base`.

2. **`Plurimath::Formatter::Numbers::Fraction#change_base`** ([lib/plurimath/formatter/numbers/fraction.rb:160-181](https://github.com/plurimath/plurimath/blob/master/lib/plurimath/formatter/numbers/fraction.rb#L160-L181)) consumes `@precision` as **count of target-base fractional digits to emit** in its `digits.times do … end` loop.

For `123.25` in base 16: decimal fractional digits = 2, but to retain 5 significant digits in hex you need 3 fractional hex digits (`7b` is 2 integer digits + `400` is 3 fractional = 5 total). Reusing the decimal count produces `7b.40` (4 sig digits).

## Suggested fix

When `:base` is non-decimal and `:significant` is provided, derive precision as `max(0, significant - integer_part_in_target_base.length)`. The isodoc-side implementation can be copied essentially verbatim (see `num_precision_from_significant` in `lib/isodoc/presentation_function/math.rb` of metanorma/isodoc, landed alongside [metanorma/isodoc#788](https://github.com/metanorma/isodoc/issues/788)):

```ruby
# Skip when base is decimal, or significant unset, or input is not E-notation
def precision_from_significant(num_string, significant, base)
  return nil if base == 10
  return nil if significant.nil?
  return nil unless /\.(?=\d+e)/.match?(num_string)

  m = num_string.match(/\A-?0?\.(\d+)e(-?\d+)\z/)
  return nil unless m
  return 0 if m[1].length <= m[2].to_i  # source is integer

  val = BigDecimal(num_string).abs
  integer_len = val.to_i.zero? ? 0 : val.to_i.to_s(base).length
  [significant - integer_len, 0].max
end
```

Behavior table this produces (verified end-to-end against Plurimath when fed back as explicit `precision:`):

| Input | Base | significant | Expected output |
|---|---|---|---|
| `0.12325e3` (=123.25) | 16 | 5 | `0x7b.400` |
| `0.123e3` (=123) | 16 | 2 | `0x7b` |
| `0.31e2` (=31) | 16 | 2 | `0x1f` |
| `0.1230e3` (=123.0) | 16 | 4 | `0x7b.00` |
| `0.25e0` (=0.25) | 16 | 2 | `0x0.40` (note: leading-`0` integer is not significant; treat its length as 0) |

Note on the `0.25e0` row: when the integer part of the value is zero, the leading "0" before the decimal point is not a significant digit, so its length should not be subtracted from `significant`. The helper detects this via `val.to_i.zero?` and uses `0` for `integer_len`.

Wiring this into `parse_number` / partition flow so it acts as a fallback when `:precision` is not supplied — ahead of the existing `precision_from(number)` fallback — should let `{base: 16, significant: 5}` produce `0x7b.400` without callers having to compute precision themselves.

## Note on broader concern

Even when `:significant` is absent, decimal-precision-as-target-base-precision is arguably wrong for non-decimal bases. The focused fix above only handles the `:significant`-present path; the broader question of how to default precision per-base is worth a separate consideration.

## Cross-references

- Downstream report: https://github.com/metanorma/isodoc/issues/788
- isodoc workaround: `lib/isodoc/presentation_function/math.rb#num_precision_from_significant` (landed alongside #788, Apr 2026)

suleman-uzair:
This PR fixes number formatting when converting values to another base with significant digits enabled. It preserves the expected trailing zeros for fractional values and adds regression specs for the reported issue.

closes #417

### DIFF
--- a/lib/plurimath/formatter/numeric_formatter.rb
+++ b/lib/plurimath/formatter/numeric_formatter.rb
@@ -127,6 +127,8 @@ module Plurimath
       # When precision is omitted, infer the target-base fractional digits
       # needed to satisfy :significant. Cap the count to the source significant
       # digits so base conversion does not invent precision for values like 0.1.
+      # Add one extra digit when the source has more significant digits so
+      # Significant can still perform the final rounding pass.
       def significant_base_precision(num, format)
         base = format[:base] || Formatter::NumberFormatter::DEFAULT_BASE
         return unless target_base?(base)
@@ -134,19 +136,17 @@ module Plurimath
         significant = format[:significant].to_i
         return if significant.zero?
 
-        decimal_precision = source_fractional_digits(num)
-        return 0 if decimal_precision.zero?
+        return 0 unless source_fractional?(num)
 
-        effective_significant = [
-          significant,
-          source_significant_digits(num),
-        ].min
+        source_significant = source_significant_digits(num)
+        effective_significant = [significant, source_significant].min
         target_precision = [
           effective_significant - target_base_integer_length(num, base),
           0,
         ].max
 
-        [decimal_precision, target_precision].max
+        target_precision += 1 if source_significant > effective_significant
+        target_precision
       end
 
       def target_base?(base)
@@ -154,11 +154,11 @@ module Plurimath
           base != Formatter::NumberFormatter::DEFAULT_BASE
       end
 
-      def source_fractional_digits(num)
+      def source_fractional?(num)
         mantissa, exponent = num.to_s.downcase.split("e", 2)
         fraction_length = mantissa.split(".", 2)[1].to_s.length
 
-        [fraction_length - exponent.to_i, 0].max
+        fraction_length > exponent.to_i
       end
 
       def source_significant_digits(num)
--- a/lib/plurimath/formatter/numeric_formatter.rb
+++ b/lib/plurimath/formatter/numeric_formatter.rb
@@ -106,18 +106,68 @@ module Plurimath
         @e = format[:e]&.to_sym || :e
         @times = format[:times]&.to_sym || "\u{d7}"
         @notation = format[:notation]&.to_sym || nil
-        @precision = update_precision(string, precision)
+        @precision = update_precision(string, precision, format)
         @exponent_sign = format[:exponent_sign]&.to_sym || nil
       end
 
-      def update_precision(num, precision)
+      def update_precision(num, precision, format)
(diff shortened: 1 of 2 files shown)


ronaldtse (approved):

suleman-uzair:
@ronaldtse Just released version `0.10.3`.

cc: @opoudjis

PR #420, opened 2026-05-07
Handle `nil` number formatter in Opal

suleman-uzair:
Handles `nil` number formatters explicitly before checking formatter capabilities.

In Opal, nil can report support for `Kernel#format` differently than MRI, which causes number formatting to dispatch into `Kernel#format` and fail in generated JavaScript bundles. This keeps the MRI behavior intact while avoiding the Opal runtime failure.

### DIFF
--- a/lib/plurimath/math/number.rb
+++ b/lib/plurimath/math/number.rb
@@ -94,6 +94,8 @@ module Plurimath
 
       def format_value_with_options(options)
         formatter = options[:formatter]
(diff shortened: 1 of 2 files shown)


ronaldtse:
Merged. Expanded nil-formatter spec coverage to `to_latex`, `to_html`, and `to_unicodemath` in addition to the original `to_asciimath` spec.

PR #422, opened 2026-05-11
Add configuration and deprecation foundation

### LINKED ISSUE #421: Add `configuration` and `deprecation` foundation
## Background

**Plurimath** has `formatter` behavior that needs to evolve without forcing immediate breaking changes. The planned `NumberFormatter` rewrite is the main driver, but the supporting API should be introduced separately so later `formatter` changes can stay smaller and easier to review.

This issue introduces two foundations: gem-level `configuration` and centralized `deprecation` handling.

## Why Configuration Is Useful

**Plurimath** currently supports passing `formatter` behavior through per-call options:

```ruby
formula.to_latex(formatter: formatter)
```

That should continue to work and should keep precedence over any configured `formatter.`

A configuration-level `formatter` is still useful because it lets callers define a shared `formatter` once and use it automatically during number rendering, without explicitly passing `formatter:` on every conversion call.

Initial API:

```ruby
Plurimath.configure do |config|
  config.number_formatter = formatter
end
```

Expected `formatter` precedence:

- If `formatter:` is passed for a conversion, that `formatter` is used.
- If no `formatter` is passed, `config.number_formatter` is used as the fallback.
- If neither is available, current behavior is preserved.

The first implementation should keep this narrow: `Math::Number` uses `config.number_formatter` only when conversion options do not already include `:formatter`.

## Why Deprecation Support Is Needed

The **NumberFormatter** rewrite will likely need to phase out or change some legacy options and behaviors. Those changes should not happen silently, and `deprecation` warnings should not be scattered as ad hoc `warn` calls.

A centralized `deprecation` mechanism gives **Plurimath** one consistent, **Opal-safe** way to announce behavior planned for removal or semantic change. It also lets consumers choose how strict they want upgrades to be.

Initial API:

```ruby
Plurimath.configure do |config|
  config.deprecation.behavior = :warn
end
```

Supported behaviors:

- `:warn` warns once per deprecated feature
- `:raise` raises a **Plurimath**-specific deprecation error
- `:silence` suppresses deprecation output

Deprecation behavior should belong to `Plurimath::Deprecation`; `Configuration#deprecation` only exposes that module through the configuration API.

## Scope

- Add `Plurimath.configuration` and `Plurimath.configure`.
- Add `Configuration#number_formatter`.
- Use `number_formatter` as a fallback in `Math::Number` only.
- Add `Plurimath::Deprecation` with `:warn`, `:raise`, and `:silence` behaviors.
- Add **Plurimath**-specific `configuration`/`deprecation` errors under the existing error structure.
- Make existing **Plurimath**-specific errors inherit from a shared `Plurimath::Error`.

## Related

This is groundwork for the full **NumberFormatter** rewrite.

suleman-uzair:
This PR adds generic `configuration` and `deprecation` foundation support for **Plurimath.**

This PR adds:
- `Plurimath.configuration` and `Plurimath.configure`
- `Configuration#number_formatter` as a configurable default formatter
- formatter precedence where explicit `formatter:` options still win over configured defaults
- centralized, Opal-safe deprecation handling with `:warn`, `:raise`, and `:silence`

This does not deprecate any existing behavior yet; it only adds the shared foundation for future configuration defaults and planned deprecation paths.

closes #421

### DIFF
--- a/lib/plurimath.rb
+++ b/lib/plurimath.rb
@@ -8,6 +8,11 @@ require "plurimath/xml_engine"
 module Plurimath
   autoload :Asciimath, "plurimath/asciimath"
   autoload :Cli, "plurimath/cli" unless RUBY_ENGINE == "opal"
+  autoload :Configuration, "plurimath/configuration"
+  autoload :Deprecation, "plurimath/deprecation"
+  autoload :Error, "plurimath/errors/error"
+  autoload :ConfigurationError, "plurimath/errors/configuration_error"
+  autoload :DeprecationError, "plurimath/errors/deprecation_error"
   autoload :Formatter, "plurimath/formatter"
   autoload :Html, "plurimath/html"
   autoload :Latex, "plurimath/latex"
@@ -28,7 +33,15 @@ module Plurimath
     Mml::V4::Configuration.adapter = adapter unless Mml::V4::Configuration.adapter
   end
 
-  module_function :mml_adapter
+  def configuration
+    @configuration ||= Configuration.new
+  end
+
+  def configure
+    yield(configuration)
+  end
+
+  module_function :mml_adapter, :configuration, :configure
 end
 
 default_adapter =
--- a/lib/plurimath/configuration.rb
+++ b/lib/plurimath/configuration.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Plurimath
+  class Configuration
+    attr_accessor :number_formatter
+
+    def deprecation
+      Deprecation
+    end
+  end
+end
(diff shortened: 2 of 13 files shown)


ronaldtse:
Merged. Key changes beyond the original contribution:
- Redesigned Deprecation from `Kernel.warn` (:warn) to collect-based (:collect) model — library never writes to output
- Added `Deprecation.notices` and `Deprecation.clear!` for API callers
- Enriched `DeprecationError` with `severity`, `feature`, `replacement`, `since`, `remove_in`, `to_s`
- Added nil-formatter guard from PR #420
- Removed all doubles/mocks from specs, replaced with behavioral tests

PR #423, opened 2026-05-11
Rewrite NumberFormatter pipeline

### LINKED ISSUE #270: Same decimal and group symbols in Significant digits
````
Plurimath::NumberFormatter.new(:en).localized_number('327428.7432878432992', precision: nil, format: {:fraction_group_digits=>2, :fraction_group=>"'", :decimal=>",", :notation=>:basic, :significant=>9}) == "327,42'8"
````

...?! This should be "327,428,74'3". Admittedly we now have an ambiguity with both group and decimal being comma, so I'll change the test, and that is an edge case; but I don't think the code should get confused by that.

_Originally posted by @opoudjis in https://github.com/plurimath/plurimath/issues/268#issuecomment-2220806549_

### LINKED ISSUE #306: Why is NumberFormatter and NumericFormatter separate? (and how to customize NumberFormatter?)
NumericFormatter is only used by NumberFormatter.

For simplicity why not combine them?

Screenshot 2024-10-06 at 6 35 56 PM

I am thinking in the context of overriding `Plurimath::NumberFormatter::Standard`.

For example, I want a number format that applies contextually (e.g. depends on where the number is, I want to format it differently).

Or for example, a year of 4 digits should be formatted as a simple string, but other numbers are formatted with digit groupings.

suleman-uzair:
This PR rewrites the **NumberFormatter** internals around structured `Source` and `Parts` data, so numeric transformations happen before localized rendering.

This PR updates:
- number parsing through `Source` and `Parts`
- significant digits, digit count, precision, base conversion, notation, and sign handling through focused formatter helpers
- final integer/fraction rendering through `PartsRenderer`
- formatter option handling through a per-call `FormatOptions` object
- locale/custom symbol resolution without mutating shared locale defaults
- custom formatter dispatch to prefer `format_number(formula, number)` while preserving existing `format` and `localized_number` behavior

Depends on #422
closes #270
closes #306


ronaldtse:
Merged (squash). Key changes beyond the original contribution:
- Rebased onto main (with #420 and #422)
- Resolved nil-formatter guard conflict
- Replaced `double()` with real `PrecisionResolver` in format_options spec

Architecture: clean pipeline (Source → Parts → NumberRenderer → renderers), immutable Parts, no shared state mutation.

PR #456, opened 2026-07-10
Dissolve MathML/OMML token resolution from `Utility`

suleman-uzair:
This PR dissolves the old shared MathML/OMML token resolver out of `Plurimath::Utility` and replaces it with explicit MathML and OMML utility entry points. It preserves the representation-aware behavior from #454 while moving format-specific token and `mrow` logic to the format owners.

## Changes

- Add `Plurimath::Mathml::Utility.resolve_token` and wire the MathML utility autoload.
- Add a private `Mathml::FunctionTokenResolver` shared only by MathML and OMML utility owners, keeping function-token semantics out of generic `Utility`.
- Add OMML-specific `resolve_text_token`, `resolve_symbol_token`, and `populate_function_classes` entry points.
- Move the OMML mrow function-population passes into `Omml::Utility` as private helpers, removing the fake `lang:` / `omml:` plumbing and the dead `under:` branch.
- Replace MathML and OMML production call sites in formula transformations and the OMML translator.
- Move accent parameter assignment into the existing OMML `accent_value` helper.
- Delete `Utility.mathml_unary_classes`, `Utility.word_resolvable?`, and generic `Utility.text_classes`.
- Move direct resolver specs out of `spec/plurimath/utility_spec.rb` into MathML/OMML utility specs, with parser integration coverage for the replaced paths.

No output behavior change relative to #454; this is a boundary-cleanup refactor that removes the flag-driven resolver and leaves MathML/OMML token handling with the owning format modules.

Depends on #455


lib/plurimath/mathml/utility.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Plurimath
+  class Mathml
+    module FunctionTokenResolver
```
ronaldtse, line 5: Quite weird to have a module inside a class.

ronaldtse (approved):

PR #480, opened 2026-08-27
Drop the control characters XML cannot represent when dumping

### LINKED ISSUE #479: xml_engine_spec: Ox.dump raises '\#x01' is not a valid XML character (ox 2.14.29)
## Summary

`spec/plurimath/xml_engine_spec.rb` — "Ox behaves like all engines `.dump` dumps simple document" — fails on current master (00c5278) with Ox 2.14.29:

```
Failure/Error: ::Ox.dump(data.xml_nodes, **options)

Ox::SyntaxError:
  '\#x01' is not a valid XML character.
# ./lib/plurimath/xml_engine/ox_engine.rb:15:in 'dump'
# ./spec/plurimath/xml_engine_spec.rb:97
```

17193 examples, 1 failure.

## Environment

- plurimath master (00c5278, 2026-07-17)
- ox 2.14.29
- lutaml-model 0.8.19 (also verified with the lutaml-model main branch — identical failure, so this is not a lutaml-model regression)

## Repro

```bash
git clone https://github.com/plurimath/plurimath && cd plurimath
bundle install
bundle exec rspec spec/plurimath/xml_engine_spec.rb
# 42 examples, 1 failure
```

## Notes

This surfaces in the lutaml-model dependent-gems CI (which tests plurimath@master against lutaml-model PRs); it started after the last green run there (2026-08-12), pointing at either a recent Ox release tightening dump validation for control characters (`\x01` is not a legal XML 1.0 character) or a recent plurimath/engine change that now lets a raw control character reach `Ox.dump`. Worth checking whether the engine should sanitize/escape control characters on dump, or whether an Ox pin is needed.

suleman-uzair:
Fixes #479.

Both XML engines write characters XML cannot carry. XML 1.0 admits only tab, newline and carriage return below `0x20`, and no character reference makes the rest legal. `Oga::Dumper` emitted `&#x000N;` for them and Ox wrote the raw byte, so a formula holding one produced a document libxml refuses to read. ox 2.14.29 surfaced this by raising `Ox::SyntaxError` rather than writing the character, which leaves the suite red on `main` until this lands.

## Changes

- Pass `invalid_replace: ""` to `Ox.dump`, placed after `**options` so a caller cannot reinstate the characters.
- Split the C0 branch in `Oga::Dumper.entities`: tab keeps the character reference it has always been written as, and the controls XML cannot represent are dropped. Nothing changes for any character XML admits.
- Cover the whole `0x00..0x1f` range in text and in attributes, so a dropped character and a wrong one both fail.

No ox pin is needed — passed at the dump site, `invalid_replace` is honoured back to ox 2.14.23.

Invalid XML can still reach output by routes this cannot see: a control character parsed into the literal text `&#x1;`, and `U+FFFE`/`U+FFFF`. Both are pre-existing on `main` and deliberately untouched here.


ronaldtse:
@suleman-uzair — a nudge to land this one: it is now the last red in the unitsml release chain.

Since metanorma/ci#414 fixed the org-wide dependent-gems 403 (the gemrc GitHub-Packages-source bug), unitsml-ruby's `dependent-gems-test` runs for real again, and its plurimath job fails on exactly the bug this PR fixes:

```
rspec ./spec/plurimath/xml_engine_spec.rb[1:1:1:2:1]
# Plurimath::XmlEngine Ox behaves like all engines .dump dumps simple document
Ox::SyntaxError: '\#x01' is not a valid XML character.
# ./lib/plurimath/xml_engine/ox_engine.rb:15:in 'dump'
```

That is 1 failure out of 17193 examples, reproduced standalone with ox 2.14.28+ (strict C0-control rejection).

I verified this branch locally against ox 2.14.29: the previously-failing example passes, and the full `spec/plurimath/xml_engine_spec.rb` is 52/52 green.

The red dependent checks on this PR predate the metanorma/ci fix and should pass on a re-run. Could this be marked ready for review and merged? Until it lands, unitsml-ruby's dependent CI stays red on this single spec.

